PR #21, opened 2017-03-03
Add signals package

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/21)

### DIFF
--- a/signals/signals.go
+++ b/signals/signals.go
@@ -1,0 +1,68 @@
+package signals
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"runtime/pprof"
+	"syscall"
+)
+
+// ContextWithShutdown returns a context that is cancelled when the SIGTERM or SIGINT signal is received.
+func ContextWithShutdown(ctx context.Context) (context.Context, context.CancelFunc) {
+	return CancelOnSignalsContext(ctx, syscall.SIGTERM, syscall.SIGINT)
+}
+
+// CancelOnSignalsContext returns a context that is cancelled when any of the provided signals are received.
+func CancelOnSignalsContext(ctx context.Context, sig ...os.Signal) (context.Context, context.CancelFunc) {
+	newCtx, cancel := context.WithCancel(ctx)
+
+	// Use a buffer of 1 in case we are not ready when the signal arrives
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, sig...)
+
+	go func() {
+		<-signals
+		cancel()
+	}()
+
+	return newCtx, cancel
+}
+
+// RegisterStackTraceWriter starts a goroutine that listens for the SIGQUIT (kill -3) signal and writes a
+// pprof-formatted snapshot of all running goroutines when the signal is received. Returns a function that unregisters
+// the listener when called.
+func RegisterStackTraceWriter(out io.Writer) (unregister func()) {
+	return RegisterStackTraceWriterOnSignals(out, syscall.SIGQUIT)
+}
+
+// RegisterStackTraceWriterOnSignals starts a goroutine that listens for the specified signals and writes a pprof-formatted
+// snapshot of all running goroutines to out when any of the provided signals are received. Returns a function that
+// unregisters the listener when called.
+func RegisterStackTraceWriterOnSignals(out io.Writer, sig ...os.Signal) (unregister func()) {
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, sig...)
+
+	cancel := make(chan bool, 1)
+	unregister = func() {
+		cancel <- true
+	}
+
+	go func() {
+		for {
+			select {
+			case <-signals:
+				err := pprof.Lookup("goroutine").WriteTo(out, 2)
+				if err != nil {
+					fmt.Fprintln(out, "Failed to dump goroutines")
+				}
+			case <-cancel:
+				return
+			}
+		}
+	}()
+
+	return unregister
+}
--- a/signals/signals_test.go
+++ b/signals/signals_test.go
@@ -1,0 +1,70 @@
+package signals_test
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/palantir/pkg/signals"
+)
+
+func TestCancelOnSignalsContext(t *testing.T) {
+	ctx, _ := signals.CancelOnSignalsContext(context.Background(), syscall.SIGHUP)
+	timer := time.NewTimer(time.Second * 3)
+
+	proc, err := os.FindProcess(os.Getpid())
+	require.NoError(t, err)
+
+	go func() {
+		err = proc.Signal(syscall.SIGHUP)
+		require.NoError(t, err)
+	}()
+
+	done := false
+	select {
+	case <-ctx.Done():
+		done = true
+	case <-timer.C:
+	}
+
+	assert.True(t, done)
+}
+
+func TestRegisterStackTraceWriterOnSignals(t *testing.T) {
+	out := &bytes.Buffer{}
+	signals.RegisterStackTraceWriterOnSignals(out, syscall.SIGHUP)
+
+	proc, err := os.FindProcess(os.Getpid())
+	require.NoError(t, err)
+	err = proc.Signal(syscall.SIGHUP)
+	require.NoError(t, err)
+
+	// add sleep because write to buffer happens on a separate channel
+	time.Sleep(1 * time.Second)
+
+	// output stack should contain current routine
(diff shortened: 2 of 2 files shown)


nmiyake:
Basically a port of our internal package, but changes some naming, removes dependency on logging and stack trace, and has some other minor differences. Looking for feedback on both content and naming/API.

jonsyu1:
Thanks for porting this! I left naming and API feedback and I'm interested to hear your thoughts.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[signals/signals.go, line 19 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeKvcCLMEntniaJKXRd:-KeKvcCLMEntniaJKXRe:bosl1j7) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals.go#L19)):*
> ```Go
> 
> // NewCancelOnSignalsContext returns a context that is cancelled when any of the provided signals are received.
> func NewCancelOnSignalsContext(sig ...os.Signal) context.Context {
> ```

This function diverges from the `context` API by removing the option to inject an underlying context. This is overly restrictive because most requests start before we register the signal context.

How about `func ContextWithSignals(ctx context.Context, signals ...os.Signal) (context.Context, cancel func())`?

If you feel strongly about having a shortcut for `context.Background()`, we can also introduce a `BaseContextWithSignals` function like net/http does [here](https://golang.org/src/net/http/h2_bundle.go?h=func.%2A%5C%28.%2A%5C%29.%2AContext#L2080).

I think it's important to expose the cancel function to the caller because we want to reclaim the resources associated with the context if the user no longer needs it.

I also think that the name `NewCancelOnSigTermSigIntContext` can be improved. I'm all for explicit names, but I think any caller that is this pedantic should just pass in the exact signals they wish to register (`SIGTERM` and `SIGINT` in this case). This function should instead be named at a higher level of abstraction that describes when it should be used, like `ContextWithShutdown` that also accepts an underlying context. What do you think?

---

*[signals/signals.go, line 36 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeL-11i0sRMIHyzxEyo:-KeL-11jC3TO-xAyL2cE:bphthra) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals.go#L36)):*
> ```Go
> // RegisterStackTraceWriterOnSigQuit starts a goroutine that listens for the SIGQUIT (kill -3) signal and writes a
> // pprof-formatted snapshot of all running goroutines when the signal is received.
> func RegisterStackTraceWriterOnSigQuit(out io.Writer) {
> ```

Same feedback as above: how about `RegisterStackTraceWriter` and  the below as `RegisterStackTraceWriterOnSignals`?

---

*[signals/signals.go, line 42 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeL-Wy2t6gvf1uQLoeh:-KeL-Wy3vU-uVu0SGWHl:b-6ngz3v) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals.go#L42)):*
> ```Go
> // RegisterStackTraceWriter starts a goroutine that listens for the specified signals and writes a pprof-formatted
> // snapshot of all running goroutines to out when any of the provided signals are received.
> func RegisterStackTraceWriter(out io.Writer, sig ...os.Signal) {
> ```

These seem fine for their use case, but it's strange that they return no control to the caller: how does a caller cancel the stack trace writer?

---

*[signals/signals_test.go, line 51 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeL0AzZzRsr6XMt2esh:-KeL0AzZzRsr6XMt2esi:b-z15fas) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals_test.go#L51)):*
> ```Go
> 
> 	// output stack should contain current routine
> 	assert.Contains(t, out.String(), "github.com/palantir/pkg/signals_test.TestRegisterStackTraceWriter")
> ```

I'm not convinced we should include the package name in the test. Seems very brittle and needs to be updated every time we rename or port the package.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/21#-:-KeL-y_zxnhkTkdlm7PJ:blzs4io)*

nmiyake:
Yup on board with most of the feedback -- felt a lot of these things myself, but probably over-optimized on maintaining the previous API. However, these changes are additive and it's still possible to swap old implementations for this one easily, so OK with going with it.

Pushing up a commit that addresses feedback now.

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[signals/signals.go, line 19 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeKvcCLMEntniaJKXRd:-KeL2CLL9UJKmIMzdgN-:b-896fix) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals.go#L19)):*

This function diverges from the `context` API by removing the option to inject an underlying context. This is overly restrictive because most requests start before we register the signal context.

How about `func ContextWithSignals(ctx context.Context, signals ...os.Signal) (context.Context, cancel func())`?

If you feel strongly about having a shortcut for `context.Background()`, we can also introduce a `BaseContextWithSignals` function like net/http does [here](https://golang.org/src/net/http/h2_bundle.go?h=func.%2A%5C%28.%2A%5C%29.%2AContext#L2080).

I think it's important to expose the cancel function to the caller because we want to reclaim the resources associated with the context if the user no longer needs it.

I also think that the name `NewCancelOnSigTermSigIntContext` can be improved. I'm all for explicit names, but I think any caller that is this pedantic should just pass in the exact signals they wish to register (`SIGTERM` and `SIGINT` in this case). This function should instead be named at a higher level of abstraction that describes when it should be used, like `ContextWithShutdown` that also accepts an underlying context. What do you think?
</blockquote></details>

Done.

---

*[signals/signals.go, line 36 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeL-11i0sRMIHyzxEyo:-KeL2J6Rx2Fyh4v1qxox:b-896fix) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals.go#L36)):*

Same feedback as above: how about `RegisterStackTraceWriter` and  the below as `RegisterStackTraceWriterOnSignals`?
</blockquote></details>

Done.

---

*[signals/signals.go, line 42 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeL-Wy2t6gvf1uQLoeh:-KeL4RWk6CaGc1mAhqBL:b-896fix) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals.go#L42)):*

These seem fine for their use case, but it's strange that they return no control to the caller: how does a caller cancel the stack trace writer?
</blockquote></details>

Done.

---

*[signals/signals_test.go, line 51 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeL0AzZzRsr6XMt2esh:-KeL4_D_ULUCkn9DaDpp:b-kgb6ks) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals_test.go#L51)):*

I'm not convinced we should include the package name in the test. Seems very brittle and needs to be updated every time we rename or port the package.
</blockquote></details>

Don't think this is a major issue, but removed package path anyway. I don't think that it's worth the overhead/effort to do things like determine test function name via reflection or anything like that -- I think that package renaming/refactoring should be rare enough that this isn't a major deal, and if it does happen, break is pretty immediate and trivial to fix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/21#-:-KeL57-tlooXHuE3dJ0i:b-piiyn7)*

jonsyu1:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---

*[signals/signals_test.go, line 51 at r1](https://reviewable.io:443/reviews/palantir/pkg/21#-KeL0AzZzRsr6XMt2esh:-KeLKOq0Sy_LYCoE337z:b-ryi3no) ([raw file](https://github.com/palantir/pkg/blob/22c77c99fb67e8f98735942be3e6ba0cb41bfaf0/signals/signals_test.go#L51)):*

Don't think this is a major issue, but removed package path anyway. I don't think that it's worth the overhead/effort to do things like determine test function name via reflection or anything like that -- I think that package renaming/refactoring should be rare enough that this isn't a major deal, and if it does happen, break is pretty immediate and trivial to fix.
</blockquote></details>

Yup, I agree.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/21#-:-KeLKRoilfDPC4-T4hCG:bnfp4nl)*

PR #27, opened 2017-03-04
Add encryption package

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/27)

### DIFF
--- a/encryption/aes.go
+++ b/encryption/aes.go
@@ -1,0 +1,74 @@
+package encryption
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"fmt"
+)
+
+type Plaintext string
+type Ciphertext string
+
+const nonceSize = 32
+
+// EncryptSym encrypts the provided plaintext with the provided symmetric key. Compatible with
+// https://github.com/palantir/encrypted-config-value.
+func EncryptSym(plain Plaintext, sym *SymmetricKey) (Ciphertext, error) {
+	gcm, err := newBlockCipher(sym)
+	if err != nil {
+		return "", err
+	}
+
+	// Generate random nonce (javax crypto calls this the initialization vector)
+	nonce, err := RandomBytes(gcm.NonceSize())
+	if err != nil {
+		return "", fmt.Errorf("Failed to generate nonce: %v", err)
+	}
+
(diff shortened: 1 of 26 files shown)


jonsyu1:
Looks good overall: just one small perf improvement suggestion: feel free to tackle in this or a separate PR.

---

Reviewed 26 of 26 files at r1.

---

*[encryption/symmetric.go, line 46 at r1](https://reviewable.io:443/reviews/palantir/pkg/27#-KeUi7WurxnX7a0hNbOZ:-KeUi7WurxnX7a0hNbO_:b-md6rah) ([raw file](https://github.com/palantir/pkg/blob/7cd908d63460ad13e76bc23e25ce1a576dc80edd/encryption/symmetric.go#L46)):*
> ```Go
> 		return NoSymmetricKey
> 	}
> 	prefix := []byte(prefixString)
> ```

Looks like  the we use `prefixString` as a byte slice more often than as a string. If we convert the type of our `aesKeyPrefix` to `[]byte` (will need to change from const to var), we can avoid the extra allocation and copy with every call to `Encode` and `Decode`. Be sure to update the error message in `Decode` to convert to string when constructing the error message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/27#-:-KeUikRqMYJkVpDNG0rj:birhraw)*

nmiyake:

---

*[encryption/symmetric.go, line 46 at r1](https://reviewable.io:443/reviews/palantir/pkg/27#-KeUi7WurxnX7a0hNbOZ:-KeVO4OUEste1sRnay2O:b-y10vfl) ([raw file](https://github.com/palantir/pkg/blob/7cd908d63460ad13e76bc23e25ce1a576dc80edd/encryption/symmetric.go#L46)):*

Looks like  the we use `prefixString` as a byte slice more often than as a string. If we convert the type of our `aesKeyPrefix` to `[]byte` (will need to change from const to var), we can avoid the extra allocation and copy with every call to `Encode` and `Decode`. Be sure to update the error message in `Decode` to convert to string when constructing the error message.
</blockquote></details>

Thanks, makes sense -- will do this in a separate PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/27)*

PR #29, opened 2017-03-05
Treat encryption key prefixes as []byte instead of string

nmiyake:
Improves performance by avoiding extra allocations.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/29)


nmiyake:
@jonsyu1 lower priority than other things going on, but pinging so this is on your radar.

PR #38, opened 2017-04-08
Improve safejson.FromYAML function

nmiyake:
* Rename to safejson.FromYAMLValue to make it clear that the input
  is a YAML value rather than the YAML itself
* Improve function documentation
* Add test to verify that function does not handle the case of
  maps in struct fields

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/38)

### DIFF
--- a/safejson/gocd_imports.json
+++ b/safejson/gocd_imports.json
@@ -20,6 +20,14 @@
             ]
         },
         {
+            "path": "github.com/palantir/pkg/vendor/github.com/stretchr/testify/require",
+            "numGoFiles": 5,
+            "numImportedGoFiles": 15,
+            "importedFrom": [
+                "github.com/palantir/pkg/safejson_test"
+            ]
+        },
+        {
             "path": "github.com/palantir/pkg/vendor/gopkg.in/yaml.v2",
             "numGoFiles": 13,
             "numImportedGoFiles": 0,
--- a/safejson/yaml.go
+++ b/safejson/yaml.go
@@ -10,8 +10,18 @@ import (
 	"strings"
 )
 
-// FromYAML converts nested map[interface{}]interface{} to map[string]interface{}.
-func FromYAML(y interface{}) (interface{}, error) {
+// FromYAMLValue returns a version of the provided input where all nested map[interface{}]interface{} values are
+// converted to map[string]interface{}. The input should be the representation of an object in
+// map[interface{}]interface{} form (as opposed to a string or []byte of the YAML itself). Returns an error if the
+// conversion fails because any of the object keys are not strings.
+//
+// Assumes that the input consists of only maps, slices, arrays, primitives and pointers to these types. Structs are
+// assumed to have been converted into a map representation -- if a struct value is encountered, it will be treated as a
+// primitive and left as-is (in particular, any maps in the struct will not be converted).
+//
+// Many YAML libraries unmarshal YAML content as map[interface{}]interface{}, but the Go JSON library requires JSON maps
+// to be represented as map[string]interface{}, so this function is helpful in converting the former to the latter.
+func FromYAMLValue(y interface{}) (interface{}, error) {
 	return fromYAMLValue(reflect.ValueOf(y), "")
 }
 
--- a/safejson/yaml_test.go
+++ b/safejson/yaml_test.go
@@ -2,13 +2,16 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package safejson
+package safejson_test
 
 import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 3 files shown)


jonsyu1:
:lgtm:

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[safejson/yaml.go, line 18 at r1](https://reviewable.io:443/reviews/palantir/pkg/38#-KhYf01l5OrXEymQwumU:-KhYf01l5OrXEymQwumV:b-hw6aj9) ([raw file](https://github.com/palantir/pkg/blob/e0e2c226845ef90a89448fe0b4b049d93e7147cf/safejson/yaml.go#L18)):*
> ```Go
> // conversion fails because any of the object keys are not strings.
> //
> // Assumes that the input consists of only maps, slices, arrays, primitives and pointers to these types. Structs are
> ```

nit: sentence fragment should be completed by providing a subject. How about `The input must consist of...`

---

*[safejson/yaml.go, line 22 at r1](https://reviewable.io:443/reviews/palantir/pkg/38#-KhYeramUFnHdTrJ3Uyx:-KhYeramUFnHdTrJ3Uyy:b-qq5pcf) ([raw file](https://github.com/palantir/pkg/blob/e0e2c226845ef90a89448fe0b4b049d93e7147cf/safejson/yaml.go#L22)):*
> ```Go
> // primitive and left as-is (in particular, any maps in the struct will not be converted).
> //
> // Many YAML libraries unmarshal YAML content as map[interface{}]interface{}, but the Go JSON library requires JSON maps
> ```

This advice feels strangely out of place in the function docs. Maybe a concrete example would be better?

---

*[safejson/yaml.go, line 24 at r1](https://reviewable.io:443/reviews/palantir/pkg/38#-KhYfCXKdTNA5Xcj9G8n:-KhYfCXKdTNA5Xcj9G8o:bhyrie8) ([raw file](https://github.com/palantir/pkg/blob/e0e2c226845ef90a89448fe0b4b049d93e7147cf/safejson/yaml.go#L24)):*
> ```Go
> // Many YAML libraries unmarshal YAML content as map[interface{}]interface{}, but the Go JSON library requires JSON maps
> // to be represented as map[string]interface{}, so this function is helpful in converting the former to the latter.
> func FromYAMLValue(y interface{}) (interface{}, error) {
> ```

I like this rename, but I thought `pkg` was in use by multiple projects. Will this be released as a major version bump? Do we even release versions of this library or does everyone depend on develop?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/38#-:-KhYfU6sv5LAHlf7_0nw:bnfp4nl)*

nmiyake:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[safejson/yaml.go, line 18 at r1](https://reviewable.io:443/reviews/palantir/pkg/38#-KhYf01l5OrXEymQwumU:-KhYlI_IjZ_ASY3uqHPs:bt6gxgx) ([raw file](https://github.com/palantir/pkg/blob/e0e2c226845ef90a89448fe0b4b049d93e7147cf/safejson/yaml.go#L18)):*

nit: sentence fragment should be completed by providing a subject. How about `The input must consist of...`
</blockquote></details>

OK will consider for later.

---

*[safejson/yaml.go, line 22 at r1](https://reviewable.io:443/reviews/palantir/pkg/38#-KhYeramUFnHdTrJ3Uyx:-KhYlGVEDiPSuIqq3i8F:bbyn5fk) ([raw file](https://github.com/palantir/pkg/blob/e0e2c226845ef90a89448fe0b4b049d93e7147cf/safejson/yaml.go#L22)):*

This advice feels strangely out of place in the function docs. Maybe a concrete example would be better?
</blockquote></details>

Example would probably be good -- I'll add later.

---

*[safejson/yaml.go, line 24 at r1](https://reviewable.io:443/reviews/palantir/pkg/38#-KhYfCXKdTNA5Xcj9G8n:-KhYkywg38aa3YHE_uyl:b-6f75sz) ([raw file](https://github.com/palantir/pkg/blob/e0e2c226845ef90a89448fe0b4b049d93e7147cf/safejson/yaml.go#L24)):*

I like this rename, but I thought `pkg` was in use by multiple projects. Will this be released as a major version bump? Do we even release versions of this library or does everyone depend on develop?
</blockquote></details>

We have a few projects that use `pkg`, but they are mostly smaller ones maintained by me. I would like to switch our larger projects over to it, but all of those still use our legacy shared code repo.

Because of this, I'm not concerned about this break. Eventually we should have a formalized process for this, but these changes were actually motivated by me trying to switch one of our larger projects over to this library.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/38)*

PR #39, opened 2017-04-10
Refactor encryption libraries and add encryptedconfigvalue

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/39)


nmiyake:
Our previous implementation of `encryption` and `encrypted-config-value` was pretty messy -- the different concerns (particularly encryption operations versus storage in `encrypted-config-value` format) bled across multiple different packages and the API and usage were inconsistent with the Java implementation.

Because of some underlying issues with the implementation (see https://github.com/palantir/encrypted-config-value/issues/61) there's still some indirection here, but over-all the concerns are pretty separated:

* The `encryption` package is concerned only with general ciphers. To any degree that serialization is supported/exposed, it's all completely generic raw bytes (for example, you can get the bytes for an `AES` key or the PKCS8 or PEM representation of an `RSA` key)
* The `encryptedconfigvalue` package adds the abstraction for `KeyPair`, `KeyWithAlgorithm` and `EncryptedValue`. It also adds logic for serialization to ECV format (`<alg>:<base64-key>` for keys, `${enc:<base64-content>}` for encrypted values)

The functional output/compatibility should match what we previously had. There are some Go API breaks, but I think they are for the better, and shouldn't be too bad to switch to for clients that want to try to do so.

I've put up a PR for an internal project that switches over to these packages as an example. This refactor should also make it easier to deal with future possible changes to the encrypted-config-value library (for example, if https://github.com/palantir/encrypted-config-value/pull/56 merges)

munkyshi:
Reviewed 28 of 28 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[encryptedconfigvalue/algorithm.go, line 97 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPFBu2X1hXR_X3nHBH:-KhPFBu2X1hXR_X3nHBI:b-dbsoxm) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/algorithm.go#L97)):*
> ```Go
> 
> func (a *algorithmDefinition) unmarshalKey(serializer KeySerializer, key SerializableKeyWithAlgorithm) (KeyWithAlgorithm, error) {
> 	decodedKey, err := base64.StdEncoding.DecodeString(key.encodedKey)
> ```

How do we handle the base32 encoded existing symmetric keys out there -- a lot of our current AES keys have the `AES32` prefix and are base32 encoded.

---

*[encryptedconfigvalue/keypairfile.go, line 104 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPKRlBYsx7D9BSAsmR:-KhPKRlCULesHuALajCb:b-bm2smu) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/keypairfile.go#L104)):*
> ```Go
> }
> 
> func toPrivKeyPath(pubKeyPath string) string {
> ```

This string modification here feels a bit iffy -- is the `.key`/`.private` stuff already well defined by spec and I'm just not aware of it?

---

*[encryptedconfigvalue/rsa_test.go, line 76 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPG5ItuUIyh3ssSuPp:-KhPG5ItuUIyh3ssSuPq:b-gfmqxw) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/rsa_test.go#L76)):*
> ```Go
> 	}{
> 		{
> 			"AES:s0u/5zMvOz9bd2/7QSJ0yaRpav9kgAmLh6GyXkttwC4=",
> ```

Using an `AES` key here.

---

*[encryptedconfigvalue/serializablekeywithalgorithm.go, line 32 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhP7yaiD1Pf_CDCckVs:-KhP7yajNZkp_dT-rIZU:bqbwt29) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/serializablekeywithalgorithm.go#L32)):*
> ```Go
> 	}
> 
> 	if _, ok := algorithms[AlgorithmType(parts[0])]; !ok {
> ```

nit: use `GetAlgorithm` to minimize codepaths?

---

*[encryptedconfigvalue/serializablekeywithalgorithm.go, line 55 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhP8V7N0Vme9mIzjizV:-KhP8V7OpRzp3coMo4yq:b-7dw6j6) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/serializablekeywithalgorithm.go#L55)):*
> ```Go
> func (skwa *SerializableKeyWithAlgorithm) UnmarshalJSON(data []byte) error {
> 	var serializedStringForm string
> 	if err := json.Unmarshal(data, &serializedStringForm); err != nil {
> ```

Is there any need to use `json.Unmarshal` vs `string(data)`?

---

*[encryption/aesgcm.go, line 28 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPBF03sYYOe7kp12ic:-KhPBF04fpwBZgkIGnz5:b-tic5pw) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryption/aesgcm.go#L28)):*
> ```Go
> 	k, err := randomBytes(keySizeBits / 8)
> 	if err != nil {
> 		return nil, fmt.Errorf("failed to generate symmetric key: %v", err)
> ```

nit: `symmetric` -> `aes`?

---

*[encryption/aesgcm.go, line 84 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPCqLFN-cZYIXoqCDr:-KhPCqLFN-cZYIXoqCDs:b-tic5pw) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryption/aesgcm.go#L84)):*
> ```Go
> func (k *aesGCMKey) newBlockCipher() (cipher.AEAD, error) {
> 	if len(k.k) == 0 {
> 		return nil, fmt.Errorf("symmetric key cannot be empty")
> ```

nit: `symmetric` -> `aes`?

---

*[encryption/aesgcm.go, line 88 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPCqxrbRZRLOoKIH0l:-KhPCqxrbRZRLOoKIH0m:b-tic5pw) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryption/aesgcm.go#L88)):*
> ```Go
> 	block, err := aes.NewCipher(k.k)
> 	if err != nil {
> 		return nil, fmt.Errorf("failed to construct symmetric cipher: %v", err)
> ```

nit: `symmetric` -> `aes`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/39)*

nmiyake:
Thanks for the feedback! I think there are a few other larger changes I'll make here as well, but appreciate the feedback on the points noted here.

---

Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[encryptedconfigvalue/algorithm.go, line 97 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPFBu2X1hXR_X3nHBH:-KhPKwEZ9FTxjFHK3F7g:b9sk6wp) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/algorithm.go#L97)):*
<details><summary><i>Previously, munkyshi (Andrew Shi) wrote…</i></summary><blockquote>

How do we handle the base32 encoded existing symmetric keys out there -- a lot of our current AES keys have the `AES32` prefix and are base32 encoded.
</blockquote></details>

Luckily, this shouldn't be too much of an issue -- if you look at the previous code for `symmetric.go`, the `Marshal` functions for both YAML and JSON will always write the key using base64 encoding, so we shouldn't have any such values persisted. The "Unmarshal" function does allow unmarshaling from base32, but unless a user explicitly created/pasted/saved such a value as persisted state, there shouldn't be anything long-lived out there that uses base32 encoding. We can still accept such input into Skylab in cases where we prompt the user for input, but we shouldn't ever actually serialize that way.

---

*[encryptedconfigvalue/keypairfile.go, line 104 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPKRlBYsx7D9BSAsmR:-KhPMS1mxxpGY-LEeOiE:b-v0zg2d) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/keypairfile.go#L104)):*
<details><summary><i>Previously, munkyshi (Andrew Shi) wrote…</i></summary><blockquote>

This string modification here feels a bit iffy -- is the `.key`/`.private` stuff already well defined by spec and I'm just not aware of it?
</blockquote></details>

Yup -- I'm not a big fan of how this is currently implemented, but this mirrors the Java implementation: https://github.com/palantir/encrypted-config-value/blob/develop/encrypted-config-value/src/main/java/com/palantir/config/crypto/KeyPair.java#L48

We don't have any actual clients that use this right now (in Java this code is used for the services and we don't have any of those in Go that use ECV in the same way), but for now thought it should mirror the Java implementation (or I can just remove the whole "load from file" section since we don't have clients for it right now)

---

*[encryptedconfigvalue/rsa_test.go, line 76 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPG5ItuUIyh3ssSuPp:-KhPO6kJkL0lv1geM17y:b-swro6q) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/rsa_test.go#L76)):*
<details><summary><i>Previously, munkyshi (Andrew Shi) wrote…</i></summary><blockquote>

Using an `AES` key here.
</blockquote></details>

You mean `RSA`, right? Thanks, good catch!

---

*[encryptedconfigvalue/serializablekeywithalgorithm.go, line 55 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhP8V7N0Vme9mIzjizV:-KhPOKPk1jbO9r5znTS5:bq8cf37) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/serializablekeywithalgorithm.go#L55)):*
<details><summary><i>Previously, munkyshi (Andrew Shi) wrote…</i></summary><blockquote>

Is there any need to use `json.Unmarshal` vs `string(data)`?
</blockquote></details>

I think doing the former could produce better errors on failure, but haven't actually tested that.

---

*[encryption/aesgcm.go, line 28 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPBF03sYYOe7kp12ic:-KhPOTPCd-ctEg0NulaL:b-f1vz7u) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryption/aesgcm.go#L28)):*
<details><summary><i>Previously, munkyshi (Andrew Shi) wrote…</i></summary><blockquote>

nit: `symmetric` -> `aes`?
</blockquote></details>

Yup, will do

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/39#-:-KhPOVsg3YoFVFLGGdQs:b-fm1tjq)*

gdearment:
Reviewed 16 of 28 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[encryptedconfigvalue/keypair.go, line 7 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhSZVRdpoxCgJEA28_2:-KhSZVRdpoxCgJEA28_3:b-rqtzrh) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/keypair.go#L7)):*
> ```Go
> package encryptedconfigvalue
> 
> type KeyPair interface {
> ```

Why make this an interface instead of a struct with 3 exported var members?

---

*[encryptedconfigvalue/keypairfile.go, line 104 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPKRlBYsx7D9BSAsmR:-KhS_679RPysjf5UF_TC:b-xvrw9h) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/keypairfile.go#L104)):*

Yup -- I'm not a big fan of how this is currently implemented, but this mirrors the Java implementation: https://github.com/palantir/encrypted-config-value/blob/develop/encrypted-config-value/src/main/java/com/palantir/config/crypto/KeyPair.java#L48

We don't have any actual clients that use this right now (in Java this code is used for the services and we don't have any of those in Go that use ECV in the same way), but for now thought it should mirror the Java implementation (or I can just remove the whole "load from file" section since we don't have clients for it right now)
</blockquote></details>

Why not use the convention that ssh RSA keypairs use, where the private key doesn't have a suffix, and the public key uses `.pub`?

---

*[encryption/key.go, line 8 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhSa1lUHMir7MAImJsR:-KhSa1lUHMir7MAImJsS:bz3lvor) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryption/key.go#L8)):*
> ```Go
> 
> // Key is an interface that can encrypt and decrypt data.
> type Key interface {
> ```

Seems a bit misleading for a `Key` to be able to encrypt and decrypt data, which the key alone isn't enough to do that, without specifying the algorithm. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/39)*

nmiyake:
Thanks for the feedback guys -- based on this feedback and some other developments I'm going to take another pass at this (want to change a few of the core concepts/abstractions here), so will close this PR for now.

---

Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[encryptedconfigvalue/keypair.go, line 7 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhSZVRdpoxCgJEA28_2:-KhSaThkTihd17nwePoZ:b-kyjzft) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/keypair.go#L7)):*
<details><summary><i>Previously, gdearment (Greg) wrote…</i></summary><blockquote>

Why make this an interface instead of a struct with 3 exported var members?
</blockquote></details>

I was thinking about this more yesterday and may lean in that direction (or something similar).

---

*[encryptedconfigvalue/keypairfile.go, line 104 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhPKRlBYsx7D9BSAsmR:-KhSafv6zSq4KXJI-66M:btsh3iw) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryptedconfigvalue/keypairfile.go#L104)):*
<details><summary><i>Previously, gdearment (Greg) wrote…</i></summary><blockquote>

Why not use the convention that ssh RSA keypairs use, where the private key doesn't have a suffix, and the public key uses `.pub`?
</blockquote></details>

I was matching the behavior of the Java library, but in all honesty there probably isn't a need for this entire file right now -- the Java side is very much tied to the Dropwizard layout/SLS spec, and we don't have any planned Go clients for this so I'll probably just remove this for now.

However, we would need to do this if we want to match the layout/locations that the Java ECV library uses for keys (see link above).

---

*[encryption/key.go, line 8 at r1](https://reviewable.io:443/reviews/palantir/pkg/39#-KhSa1lUHMir7MAImJsR:-KhSb99L4VuXSOBPRx_8:b-76ftiv) ([raw file](https://github.com/palantir/pkg/blob/25045823bea21e91c53e46919310119c9e7a0e19/encryption/key.go#L8)):*
<details><summary><i>Previously, gdearment (Greg) wrote…</i></summary><blockquote>

Seems a bit misleading for a `Key` to be able to encrypt and decrypt data, which the key alone isn't enough to do that, without specifying the algorithm. 
</blockquote></details>

Yup fair -- as part of the refactor I was thinking about yesterday, will probably separate this a bit further where the actual keys themselves are just the key data (`[]byte` or an alias/wrapper type around that for AES, `rsa.Priv/Pub` for RSA, etc.) and then have the algorithms take those instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/39#-:-KhSbReY8jotMuNnkp8Y:b-w65pe5)*

nmiyake:
Needs a bit more iteration locally based on feedback and other factors -- closing PR for now.

nmiyake:
Iterated on this quite a bit and feel good about where it is now; so going to go ahead and merge. Post-merge feedback still welcome if anyone sees any issues, though.

PR #40, opened 2017-04-19
add context.Int to cli context

griechel:
@nmiyake

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/40)

### DIFF
--- a/cli/context.go
+++ b/cli/context.go
@@ -42,7 +42,7 @@ func (ctx *Context) Slice(name string) []string {
 }
 
 func (ctx *Context) Int(name string) int {
-	value, err := strconv.Atoi(ctx.get(name).(string))
+	value, err := strconv.Atoi(ctx.String(name))
 	if err != nil {
 		panic(fmt.Errorf("interface conversion: interface is string, not int"))
 	}
--- a/cli/context.go
+++ b/cli/context.go
@@ -6,6 +6,7 @@ package cli
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -38,6 +39,14 @@ func (ctx *Context) Duration(name string) time.Duration {
 
 func (ctx *Context) Slice(name string) []string {
 	return ctx.get(name).([]string)
+}
+
+func (ctx *Context) Int(name string) int {
+	value, err := strconv.Atoi(ctx.get(name).(string))
+	if err != nil {
+		panic(fmt.Errorf("interface conversion: interface is string, not int"))
+	}
+	return value
 }
 
 func (ctx *Context) get(name string) interface{} {
--- a/cli/context_test.go
+++ b/cli/context_test.go
@@ -111,6 +111,14 @@ func TestTypedGetFunctionsReturnValue(t *testing.T) {
 			flagValue: []string{time.Minute.String()},
 			value:     time.Minute,
 		},
+		{
+			getFunction: func(ctx Context) interface{} {
+				return ctx.Int(flagName)
+			},
+			testFlag:  flag.StringFlag{Name: flagName, Value: "10"},
+			flagValue: []string{"5"},
+			value:     5,
+		},
 	}
 
 	for _, c := range cases {
@@ -144,6 +152,12 @@ func TestTypedGetFunctionsPanicOnMissingFlags(t *testing.T) {
 				return ctx.Duration(missingFlagName)
 			},
 			testFlag: flag.DurationFlag{Name: flagName, Value: "120s"},
+		},
+		{
+			getFunction: func(ctx Context) interface{} {
+				return ctx.Int(missingFlagName)
+			},
+			testFlag: flag.StringFlag{Name: flagName, Value: "5"},
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


nmiyake:
Minor feedback

---

Reviewed 2 of 2 files at r1.

---

*[cli/context.go, line 45 at r1](https://reviewable.io:443/reviews/palantir/pkg/40#-Ki6tvqwtJBEzaigH_JJ:-Ki6tvqxMlGM9MVFH6iD:b-2xahqk) ([raw file](https://github.com/palantir/pkg/blob/951e95a819a303709a3d5d8420123b3f437deae1/cli/context.go#L45)):*
> ```Go
> 
> func (ctx *Context) Int(name string) int {
> 	value, err := strconv.Atoi(ctx.get(name).(string))
> ```

Use `ctx.String(name)` in the call instead of duplicating the logic

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/40#-:-Ki6uBwqpZgSlWZQ5zv4:b-oh3hfv)*

cli/context.go
```diff
@@ -40,6 +41,14 @@ func (ctx *Context) Slice(name string) []string {
 	return ctx.get(name).([]string)
 }
 
+func (ctx *Context) Int(name string) int {
+	value, err := strconv.Atoi(ctx.get(name).(string))
+	if err != nil {
+		panic(fmt.Errorf("interface conversion: interface is string, not int"))
```
bmoylan: I don't really like panicking on user input, would it be too weird to return an error for Int if the others don't?

nmiyake:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[cli/context.go, line 47 at r1](https://reviewable.io:443/reviews/palantir/pkg/40#-Ki6qDANc4ng_9brbzTz-r1-47:-Ki6uzvVkLjth2Z-XiU1:b-x80thp) ([raw file](https://github.com/palantir/pkg/blob/951e95a819a303709a3d5d8420123b3f437deae1/cli/context.go#L47)):*

I don't really like panicking on user input, would it be too weird to return an error for Int if the others don't?
</blockquote></details>

Yeah, these functions are meant to assert as part of the casting -- all of the other ones panic as well.

However, you could probably argue that, if we go this route, we should introduce `Int` as a top-level flag type rather than as a side-effect of parsing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/40)*

cli/context.go
griechel: All the other ones panic too if casting goes awry

griechel:
@nmiyake addressed comments. Good point about the flag. Is that something that I can add in a later PR? This small change is blocking another project's work at the moment

bmoylan:
If you don't want to implement `flag.IntFlag` can't you just do the strconv in your application code until then? This isn't _really_ blocking...

---

Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/40#-:-Ki6xTUA3GbLu4okw103:b-vy4tu3)*

nmiyake:
Brad beat me to it -- I have the exact same feeling. `pkg` is very general-purpose and has a higher quality bar -- we want to be sure that things that are added here are conceptually consistent.

Agree that, if we want to semantically declare that a particular flag must store an integer value, then we should do the full change and add an `IntFlag` and propagate that all the way through.

If we don't do that, then your application is simply declaring by convention that it expects the content of particular flags to be `int` values. If that's the case, I would like that logic to live at the application-level rather than in this package. It's only a few lines of code and, as far as I can tell, doesn't depend on any internal/private state, so it should be possible to have on the client-side.

---

Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/40#-:-Ki6xjD0lchIOgOHQoDN:bwbymh9)*

griechel:
You're right, there is definitely a workaround. I'll do that and re-open a PR that does it right and includes the IntFlag

jonsyu1:
I agree with Brad

On Apr 19, 2017, at 1:45 PM, Brad Moylan <notifications@github.com<mailto:notifications@github.com>> wrote:


If you don't want to implement flag.IntFlag can't you just do the strconv in your application code until then? This isn't really blocking...

________________________________

Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

________________________________

Comments from Reviewable[reviewable.io]<https://urldefense.proofpoint.com/v2/url?u=https-3A__reviewable.io-3A443_reviews_palantir_pkg_40-23-2D-3A-2DKi6xTUA3GbLu4okw103-3Ab-2Dvy4tu3&d=DwMFaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=b6UG3qRebLXmvfXiVSoNxg&m=kP7mH5OgT3WINI2TdoAv2VSVma-x0IErH8LZo0gPAVg&s=frs4-H00d0qX_p9WrXvAIip5z8uOC-XhD6ctmc8DVTQ&e=>

PR #48, opened 2017-05-12
Do not transform big.Float while encoding json for perf reasons

DzmitryDulko:
Porting from palantir/go-palantir/safejson package
    transform.Rules.Apply does not properly convert *big.Float to
    json.Number: it has always returned a *big.Float. We should stop
    applying this no-op transformation, because it is exorbitantly costly.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/48)

### DIFF
--- a/safejson/marshal.go
+++ b/safejson/marshal.go
@@ -7,19 +7,16 @@
 //
 // - json.Decoder.UseNumber
 // - json.Encoder.SetEscapeHTML(false)
-// - json.Encoder.Encode(big floats as json.Number)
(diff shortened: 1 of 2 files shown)


nmiyake:
Hmm this isn't triggering on CircleCI and I can't amend. Can you create a new branch in your repo and then create a PR from that branch into `pkg/master`?

nmiyake:
Closing in favor of #49 (CircleCI build didn't kick off for this PR for some reason)

PR #56, opened 2017-06-02
Add tableprinter

nmiyake:
Differences from internal version:

* Use built-in errors instead of stacktrace
* Add "includeHeaderSeparator" functionality which allows '-' to
  be printed as a separator under headers

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/56)

### DIFF
--- a/tableprinter/tableprinter.go
+++ b/tableprinter/tableprinter.go
@@ -1,0 +1,131 @@
+// Copyright 2016 Palantir Technologies. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package tableprinter implements a pretty printer that writes rows and columns
+// as a formatted table.
+package tableprinter
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"text/tabwriter"
+)
+
+// A ColumnGetter returns the selected ColumnGetter value of a row.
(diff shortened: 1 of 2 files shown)


nmiyake:
I know the long param list more or less matches `tabwriter`, but wouldn't be opposed to using param decoration style for creating new tab writers to make API more readable/fluent (and would also reduce concerns around breaking API if we later introduce other options).

Also open to feedback on the "separator" style. I know many ASCII tables don't do this, but I personally find it helpful. Example output:

```
Name     Type      JSON Field
----     ----      ----------
Name     string    
Value    string    my-value
Tag      string    json-tag
```

jonsyu1:
:lgtm:. I like your param decoration idea.

One thought I had was allowing the caller to define custom header formats: i.e. using `=` as a separator instead of `-` or printing multi-line headers. With the param decoration pattern, we could allow callers to define their own `printHeader` function that would accept the underlying `tabwriter` and the `columns`. But if they just want the standard header, than they can use the prebuilt `tableprinter.HeaderWithSeparator`.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/56#-:-Klf-VLY2bvU-Nb8VryL:b-19llj2)*

PR #58, opened 2017-06-12
Add httpserver package

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/58)

### DIFF
--- a/httpserver/httpserver.go
+++ b/httpserver/httpserver.go
@@ -1,0 +1,113 @@
+// Copyright 2016 Palantir Technologies. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package httpserver provides functions that are useful for HTTP servers.
+package httpserver
+
+import (
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+// AvailablePort returns a best-effort determination of an available port. Does so by opening a TCP listener on
+// localhost, determining the port used by that listener, closing the listener and returning the address that was used
+// by the listener. This is best-effort because there is no way to guarantee that another process will not take the port
+// between the time when the listener is closed and the returned port is used by the caller.
+func AvailablePort() (port int, rErr error) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	defer func() {
+		if err := l.Close(); err != nil && rErr == nil {
+			rErr = err
+		}
+	}()
+	if err != nil {
+		return 0, err
+	}
+
+	addrString := l.Addr().String()
+	port, err = strconv.Atoi(addrString[strings.LastIndex(addrString, ":")+1:])
+	if err != nil {
+		return 0, err
+	}
+
+	return port, nil
+}
+
+// URLReady returns a channel that is sent "true" when an http.Get executed against the provided URL returns a response
+// with status code http.StatusOK. This is a convenience function that calls Ready with a readyCall that consists of
+// sending a GET request to the provided URL and a readyResp that returns true on a 200 status.
+func URLReady(url string, timeout time.Duration) <-chan bool {
+	return Ready(func() (*http.Response, error) {
+		return http.Get(url)
+	}, func(resp *http.Response) bool {
+		return resp.StatusCode == http.StatusOK
+	}, timeout)
+}
+
+// Ready returns a channel that is sent "true" when the provided readyCall returns a nil error. The readyCall is invoked
+// once every 100ms until it either returns a nil error or the provided timeout duration is reached, in which case
+// "false" is sent on the channel.
+//
+// Note that any call that returns a nil error for readyCall is interpreted to mean that the server is ready. For most
+// HTTP clients, any response (including a response with an error code) will result in a nil error: the error is
+// typically only non-nil if a transport-level failure occurs.
+//
+// readyCall should by a function that returns quickly. At most one readyCall will be running at a particular time.
+//
+// Example:
+//
+//   Ready(func() (*http.Response, error) {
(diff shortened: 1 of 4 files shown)


nmiyake:
@jonsyu1 for SA, but merging as-is for now.

httpserver/httpserver.go
```diff
@@ -0,0 +1,113 @@
...
+// Note that any call that returns a nil error for readyCall is interpreted to mean that the server is ready. For most
+// HTTP clients, any response (including a response with an error code) will result in a nil error: the error is
+// typically only non-nil if a transport-level failure occurs.
+//
+// readyCall should by a function that returns quickly. At most one readyCall will be running at a particular time.
+//
+// Example:
+//
+//   Ready(func() (*http.Response, error) {
+//     return http.Get(fmt.Sprintf("http://localhost:%d/example/ready", port))
```
jonsyu1: Is it necessary to separate the two steps into two separate functions? I get it's more modular this way, but it increases the complexity of the arguments this function takes.

httpserver/httpserver.go
```diff
@@ -0,0 +1,113 @@
...
+//   }, func(resp *http.Response) bool {
+//     return resp.StatusCode == http.StatusOK
+//   }, 5 * time.Second)
+func Ready(readyCall func() (*http.Response, error), readyResp func(*http.Response) bool, timeout time.Duration) <-chan bool {
```
jonsyu1: Why not accept a context instead of a timeout? It's more flexible and easier to compose.

httpserver/httpserver.go
```diff
@@ -0,0 +1,113 @@
...
+
+	ready := make(chan bool)
+	go func() {
+		timeout := time.NewTimer(timeout)
+		defer timeout.Stop()
+
+		// start a separate goroutine with the ticker. Done so that the possibly expensive action will not
+		// block the timeout.
+		go func() {
+			ticker := time.NewTicker(100 * time.Millisecond)
```
jonsyu1: It's a shame the ticker is hard-coded into the `httpserver.Ready` function.

httpserver/httpserver.go
nmiyake: Thinking was that `ReadyURL` should be the standard use for most folks, but otherwise want to provide full flexibility. Trying to have medium flexibility is harder, and although we could use something like a params approach, didn't know if that would be worth it

httpserver/httpserver.go
nmiyake: Could do that as well, but I think this is how people typically think of waiting -- otherwise, standard use will almost always have a separate "wait in goroutine and then cancel context" (which I guess could be fine as well -- maybe this should be the common/utility case and we can keep the generic one super-generic)

httpserver/httpserver.go
nmiyake: Yeah, didn't want to expose it as yet another param, but open to doing so.

nmiyake:
Should probably have clarified that the intended use case for this is in unit/integration tests (and a lot of the API decisions are based on that). Fine with either renaming package to indicate this or moving into something like our server framework to have people vendor from there.

There is the separate discussion of whether tests should even have/use "utility" packages, but some of this logic isn't totally trivial, and don't really want to copy-paste/rewrite for every test package for every web server we create.

httpserver/httpserver.go
jonsyu1: I agree that the params approach is overkill here. I'm just flagging that it feels weird that these are both passed in as callbacks when these feel like the core code that we want to run. It feels like the inversion of control has made it harder to understand what's happening.

jonsyu1:
OK thanks for the context (no pun intended). I like `URLReady` but `Ready` looks clunky to me. I would personally prefer to see the `for` `select` loop in any test that needs to wait for something other than `GET` to return `200`.

nmiyake:
I think this will change on the product side soon, but there was a consumer yesterday that used this to do a POST and wait for the result of that rather than a simple "GET" (although I think they're going to implement availability to be able to make the switch to GET).

nmiyake:
I have a PR to make the parameter configuration a bit cleaner (https://github.com/palantir/pkg/pull/64), but want to mention that I have found uses for `Ready` as well so would like to keep it. One example of this is checking to see if a server in a unit test that uses `https` has started -- in that case, it's not sufficient to use `http.Get` because the call needs to use a transport that has TLS configured properly.

jonsyu1:
I mentioned in that [PR](https://github.com/palantir/pkg/pull/64#discussion_r125354651) why I still find it strange. I think test writers would get more flexibility and less drag if they simply wrote the for and select statements themselves.

nmiyake:
Don't want to go down the full rabbit hole of unit testing versus integration testing etc., but for integration-style testing where you want a server to be up and available before you do the "real" tests (test the things that you care about), being able to check/assert that the server has started with a one-line library function call rather than implementing the logic manually actually saves a lot of time.

I would say the same of the `dirs.TempDir` function (https://github.com/nmiyake/pkg/blob/develop/dirs/testdir.go#L33) -- even though the primitives it uses are pretty simple and could easily be rewritten for every test, being able to write:

```go
tmpDir, cleanup, err := dirs.TempDir(wd, "")
defer cleanup()
require.NoError(t, err)
```

(https://github.com/palantir/godel/blob/master/apps/distgo/integration_test/integration_test.go#L42)

to create and clean up a temporary directory is a lot quicker than writing the logic manually.

I know that "A little copying is better than a little dependency." is a [Go proverb](https://go-proverbs.github.io/), but the bottom line is that I've found that I'm able to write and reason about tests much more quickly with small helper functions like these.

Happy to chat more in person, but just wanted to offer my perspective/background.

jonsyu1:
Cool, I don't have any strong objections to this: I just wanted to see your side of the argument.

PR #59, opened 2017-06-15
Add viper support to CLI flags

nmiyake:
* Add FlagValue function to cli.Context
* Add cliviper package and provide implementation of viper.FlagValueSet
  for cli.Context
* Add tests

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/59)


bmoylan:
:lgtm:

---

Reviewed 340 of 340 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/59#-:-Kmh6yE8Q-X2YMkakkBt:bg7igk9)*

nmiyake:
Note: ideally I wanted to be able to do this without adding `viper` as a dependency, but unfortunately it's needed to satisfy the `viper.FlagValueSet` interface. Additions to vendor are due to that. However, the `cli` package itself does not depend on any of these new packages (only the `cliviper` package).

The regular `cli` package also does add support for `FlagValue` so that functionality can be used even without imports, but the nicer `FlagValueSet` (which will register all flags on a provider) does seem to require the import (although if anyone knows any clever ways to satisfy the Go compiler without doing this I'm all ears)

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[cli/cliviper/cliviper.go, line 19 at r1](https://reviewable.io:443/reviews/palantir/pkg/59#-Kmh9AcjLC0TyU1jZkjZ:-Kmh9AcjLC0TyU1jZkj_:b3fmt61) ([raw file](https://github.com/palantir/pkg/blob/9164683daeaeed1952fc3708ab65630776101fda/cli/cliviper/cliviper.go#L19)):*
> ```Go
> type flagValueSet cli.Context
> 
> func (p *flagValueSet) VisitAll(fn func(viper.FlagValue)) {
> ```

Ideally, I wanted to be able to define the `FlagValueSet` interface within this package and just match signatures so that the import isn't required.

Unfortunately, duck typing doesn't work for the `VisitAll(fun(viper.FlagValue))` part -- even if you declare a `cli.FlagValue` that has the same method set as `viper.FlagValue`, `func(cli.FlagValue)` is viewed as a different type than `func(viper.FlagValue)` by the Go compiler.

`cliviper` is declared as a separate package to minimize dependencies -- this way, `cliviper` imports the `viper` package, but the regular `cli` package does not.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/59#-:-Kmh9fRKBhjnAel4Ci3A:b-2m2gxh)*

bmoylan:
Thanks for the info -- I think that's fine. If I want the cliviper functionality I almost definitely already have a viper dependency in my application, so it shouldn't bloat consumers even if it does bloat this repo a bit.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/59#-:-KmhJ5wx1ASl5OZ_jaqJ:b-bc5kxa)*

bmoylan:
@nmiyake is this good to go? I actually just ran into a use case where I need to optionally load config from flags or yaml and this seems pretty perfect for that.

nmiyake:
Yup should be good -- will merge now.

PR #62, opened 2017-06-27
Use dep to manage dependencies

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/62)


nmiyake:
No urgency here, but the format for the `dep` files has been locked down (https://github.com/golang/dep), so considering using it for public repos.

PR #64, opened 2017-07-02
Update httpserver functions

nmiyake:
* Add documentation
* Use parameter function paradigm to make more parameters optional,
  expose more configurable parameters and provide sensible default values

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/64)

### DIFF
--- a/httpserver/httpserver_test.go
+++ b/httpserver/httpserver_test.go
@@ -21,7 +21,7 @@ func TestURLReady(t *testing.T) {
 	}))
 	defer server.Close()
 
-	r := httpserver.URLReady(server.URL, 200*time.Millisecond)
+	r := httpserver.URLReady(server.URL, httpserver.WaitTimeoutParam(200*time.Millisecond))
 	assert.True(t, <-r)
 }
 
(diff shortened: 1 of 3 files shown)


httpserver/httpserver_test.go
```diff
@@ -55,11 +55,13 @@ func TestReadyPut(t *testing.T) {
 	}))
 	defer server.Close()
 
-	r := httpserver.Ready(func() (*http.Response, error) {
-		return http.Post(server.URL, "text", nil)
-	}, func(resp *http.Response) bool {
-		return resp.StatusCode == http.StatusAccepted
-	},
-		200*time.Millisecond)
+	r := httpserver.Ready(
```
jonsyu1: We talked why I'm not a huge fan of `httpserver.Ready` before, but I did not give a great example of an alternative. I found myself writing this test like so:

```
        // Configure retry conditions
	var (
		ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
		retry       = time.NewTicker(200 * time.Millisecond)
	)
	defer cancel()

	for {
		select {
		case <-ctx.Done():
			t.Fatal("Timed out waiting for server to respond")
		case <-retry.C:
			resp, err := http.Post(server.URL, "text", nil)
			if err == nil && resp.StatusCode == http.StatusAccepted {
				break
			}
		}
	}
```
What I like about this is the test gets to explicitly decide:
* what duration to retry
* how many times it retries
* when it stops retrying

all without buying into a framework for retrying. I do not see much complexity in writing a select nested in a for loop.

PR #70, opened 2017-08-29
Pass a configurable context.Context along with cli.Context

joshdk:
The basic motivation behind this PR is to add the ability to configure a custom `context.Context` on `cli.App` **once** (such as a complex service logger), and consume that context from anywhere in the cli. Specifically:

- Off of the `cli.App` itself
- Inside of `cli.App.ErrorHandler`
- Inside of `cli.App.Command.Action`
- Inside of `cli.App.Command.Subcommands`

I am aiming for the guarantee that if an app is created normally via `cli.NewApp()`, then a valid (non-`nil`) context is provided in all cases outlined above.

Hopefully this help reduce code duplication in consumer applications, and enable a more straight-forward approach with custom application logging.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/70)

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -5,6 +5,7 @@
 package cli
 
 import (
+	"context"
 	"io"
 	"os"
 
@@ -22,6 +23,7 @@ type App struct {
 	Manpage        *Manpage
 	Backcompat     []Backcompat
 	OnExit         OnExit
+	ContextConfig  func(context.Context) context.Context
 	ContextOptions []ContextOption
 }
 
--- a/cli/context.go
+++ b/cli/context.go
@@ -5,6 +5,7 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"time"
@@ -16,10 +17,16 @@ type Context struct {
 	Path       []string
 	IsTerminal func() bool
 
+	context   context.Context
+	cancel    context.CancelFunc
 	defaults  map[string]interface{}
 	specified map[string]interface{}
 	// stores all flag values in the order they were encountered
 	allVals map[string][]interface{}
+}
+
+func (ctx *Context) Context() context.Context {
+	return ctx.context
 }
 
 func (ctx *Context) Has(name string) bool {
--- a/cli/parse.go
+++ b/cli/parse.go
@@ -5,6 +5,7 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"syscall"
@@ -25,6 +26,15 @@ func (app *App) parse(args []string) (Context, error) {
 		specified: map[string]interface{}{},
 		allVals:   map[string][]interface{}{},
 	}
+
+	baseContext := context.Background()
+
+	if app.ContextConfig != nil {
+		baseContext = app.ContextConfig(baseContext)
+	}
+
+	ctx.context, ctx.cancel = context.WithCancel(baseContext)
+
 	args = args[1:] // skip name of binary
 	fillDefaults(&ctx)
 	for {
--- a/cli/run.go
+++ b/cli/run.go
@@ -136,6 +136,9 @@ func runAction(ctx Context) error {
 	go func() {
 		if _, ok := <-signals; ok {
 			once.Do(onExit)
+			if ctx.cancel != nil {
+				ctx.cancel()
+			}
 			os.Exit(1)
 		}
 	}()
--- a/cli/run_test.go
+++ b/cli/run_test.go
@@ -6,6 +6,7 @@ package cli_test
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"testing"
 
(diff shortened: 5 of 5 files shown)


bmoylan:
I think this is an acceptable exception to "never store a context" given that the cli.App is intended to handle global app configuration. Would want @nmiyake's opinion as well.

nmiyake:
Thanks for the PR! Definitely on board with the general concept, but I think it makes more sense to have this on the `cli.Context` rather than the `cli.App`.

---

Reviewed 4 of 4 files at r1.

---

*[cli/app.go, line 26 at r1](https://reviewable.io:443/reviews/palantir/pkg/70#-KskPAubZK_IJqIUPcb2:-KskPAubZK_IJqIUPcb3:bjjmi3l) ([raw file](https://github.com/palantir/pkg/blob/39b9b01b033d2265c370357439b3e3454f7f9fb8/cli/app.go#L26)):*
> ```Go
> 	Backcompat     []Backcompat
> 	OnExit         OnExit
> 	Context        context.Context
> ```

Storing the context as a public variable of `App` seems to fragile -- users will have the ability to re-assign this if they want. Also, if they create a `cli.App` directly, they will have to remember to assign the context themselves.

Would prefer to store the `context.Context` as a private variable on `cli.Context` -- the construction of that object is controlled more tightly and it is already provided to all of the commands (since it is also a context, albeit of a different kind). The `cli.Context` object itself can then expose a function that returns the stored context.

When you make this change, would also recommend wrapping it in a cancellable and hooking it up so that the context's `cancel` function is called in the exit portion of `runAction` (between `once.Do(onExit)` and `os.Exit(1)`) so that any routines that use the context will get the cancel signal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/70#-:-KskOhZmrPnGGLr_Uc5z:bpbkbz2)*

nmiyake:
Would like to add a `nil` check before calling cancel function to be extra defensive, but otherwise I think this should be good

---

Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[cli/run.go, line 139 at r2](https://reviewable.io:443/reviews/palantir/pkg/70#-KsqPCXK3DIG8jqitEU1:-KsqPCXK3DIG8jqitEU2:bblvha8) ([raw file](https://github.com/palantir/pkg/blob/d94c6b54f2cff6c4d698ede92874cfbb770a46f8/cli/run.go#L139)):*
> ```Go
> 		if _, ok := <-signals; ok {
> 			once.Do(onExit)
> 			ctx.cancel()
> ```

This should be wrapped in a `nil` check so that function doesn't panic if `ctx.cancel` is `nil` (current construction should guarantee that is not the case, but I think it's better to be defensive/overly cautious)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/70#-:-KsqPQBvynEBoz3z1zMi:bnjajzg)*

joshdk:
@nmiyake If you could take a final look at the PR, that would be much appreciated!

joshdk:
Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[cli/app.go, line 26 at r1](https://reviewable.io:443/reviews/palantir/pkg/70#-KskPAubZK_IJqIUPcb2:-KskhL40KbcXZsDzNICy:b-tcdd1f) ([raw file](https://github.com/palantir/pkg/blob/39b9b01b033d2265c370357439b3e3454f7f9fb8/cli/app.go#L26)):*

Storing the context as a public variable of `App` seems to fragile -- users will have the ability to re-assign this if they want. Also, if they create a `cli.App` directly, they will have to remember to assign the context themselves.

Would prefer to store the `context.Context` as a private variable on `cli.Context` -- the construction of that object is controlled more tightly and it is already provided to all of the commands (since it is also a context, albeit of a different kind). The `cli.Context` object itself can then expose a function that returns the stored context.

When you make this change, would also recommend wrapping it in a cancellable and hooking it up so that the context's `cancel` function is called in the exit portion of `runAction` (between `once.Do(onExit)` and `os.Exit(1)`) so that any routines that use the context will get the cancel signal.
</blockquote></details>

Updated with a refactor.

---

*[cli/run.go, line 139 at r2](https://reviewable.io:443/reviews/palantir/pkg/70#-KsqPCXK3DIG8jqitEU1:-KszcE5r2Z7cko3VRKMT:b-896fix) ([raw file](https://github.com/palantir/pkg/blob/d94c6b54f2cff6c4d698ede92874cfbb770a46f8/cli/run.go#L139)):*

This should be wrapped in a `nil` check so that function doesn't panic if `ctx.cancel` is `nil` (current construction should guarantee that is not the case, but I think it's better to be defensive/overly cautious)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/70)*

nmiyake:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/70#-:-Kt-t9QefQbUnmucaFOe:bnfp4nl)*

nmiyake:
Thanks!

PR #75, opened 2017-09-07
Update ContextConfig function to take the cli Context

chewr:
Refactor the signature of ContextConfig to take the cli Context object. This is necessary in order to decorate the `context.Context` with a logger that is configured from a config file

@joshdk FYSA since this is a backcompat break to your fix

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/75)


palantirtech:
Thanks for your interest in [palantir/pkg](https://github.com/palantir/pkg), @chewr! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

chewr:
@nmiyake fysa

nmiyake:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/75#-:-KtSKx8VfSiYm6o2WxHW:bnfp4nl)*

nmiyake:
@joshdk can you take a quick look at this as well? Technically this breaks the API, but I think it's a pretty useful change since it allows you to parse flags in configuring the context, which is important if you want to be able to use the flags that are in the `cli.Context` to configure the `context.Context` (@chewr ran into this while trying to configure a CLI app context with a logger).

joshdk:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[cli/run.go, line 47 at r1](https://reviewable.io:443/reviews/palantir/pkg/75#-KtS_PTlMHnGWi_Qrzll:-KtS_PTlMHnGWi_Qrzlm:b-l8zhsp) ([raw file](https://github.com/palantir/pkg/blob/fb355390b7eab8eb97581f70a5306898352be9b5/cli/run.go#L47)):*
> ```Go
> 	}
> 
> 	ctx.context, ctx.cancel = context.WithCancel(baseContext)
> ```

This in unrelated to *this* PR, but there is a very specific NPE risk where a consumer both implements the `app.ContextConfig` function, **and** has it return `nil`. I have filed https://github.com/palantir/pkg/issues/76 to correct this.

---

*[cli/run_test.go, line 98 at r1](https://reviewable.io:443/reviews/palantir/pkg/75#-KtSYXSmalyjTLK1LXQc:-KtSYXSmalyjTLK1LXQd:bld4lmk) ([raw file](https://github.com/palantir/pkg/blob/fb355390b7eab8eb97581f70a5306898352be9b5/cli/run_test.go#L98)):*
> ```Go
> func TestRunContext(t *testing.T) {
> 
> 	var customContextFunc = func(_ cli.Context, ctx context.Context) context.Context {
> ```

I would prefer to see a test case actually consume the contents inside of the `cli.Context` and then assert that the (flag/etc) information was propagated correctly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/75)*

nmiyake:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[cli/run.go, line 47 at r1](https://reviewable.io:443/reviews/palantir/pkg/75#-KtS_PTlMHnGWi_Qrzll:-KtSdDOFO1hrLgM1_zH1:bd9oyl7) ([raw file](https://github.com/palantir/pkg/blob/fb355390b7eab8eb97581f70a5306898352be9b5/cli/run.go#L47)):*
<details><summary><i>Previously, joshdk (Josh Komoroske) wrote…</i></summary><blockquote>

This in unrelated to *this* PR, but there is a very specific NPE risk where a consumer both implements the `app.ContextConfig` function, **and** has it return `nil`. I have filed https://github.com/palantir/pkg/issues/76 to correct this.
</blockquote></details>

Merged https://github.com/palantir/pkg/pull/77 to fix this issue. Update base of this PR and copy new code to resolve this issue.

---

*[cli/run_test.go, line 98 at r1](https://reviewable.io:443/reviews/palantir/pkg/75#-KtSYXSmalyjTLK1LXQc:-KtSdLBaemqOoYOqS474:b-32w0iv) ([raw file](https://github.com/palantir/pkg/blob/fb355390b7eab8eb97581f70a5306898352be9b5/cli/run_test.go#L98)):*
<details><summary><i>Previously, joshdk (Josh Komoroske) wrote…</i></summary><blockquote>

I would prefer to see a test case actually consume the contents inside of the `cli.Context` and then assert that the (flag/etc) information was propagated correctly.
</blockquote></details>

Sounds reasonable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/75)*

cli/run_test.go
```diff
@@ -94,9 +95,20 @@ func TestRunErrorHandler(t *testing.T) {
 }
 
 func TestRunContext(t *testing.T) {
+	testFlagName := "test-flag"
```
nmiyake: Fine for this PR, but FYI you can have const blocks in functions as well, so this can be rewritten as:

```go
const(
    testFlagName = "test-flag"
    testFlagValue = "foobar"
)
```

nmiyake:
:lgtm:

---

Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/pkg/75#-:-KtT3oBwhLGpwAGTpJQu:bnfp4nl)*

Issue #76, opened 2017-09-07
Check for nil return from custom app.ContextConfig function


joshdk:
There exists a very specific NPE risk where a consumer of `github.com/palantir/pkg/cli` both implements a custom `app.ContextConfig` function, **and** has it return `nil`.

[This call to `context.WithCancel`](https://github.com/palantir/pkg/blob/master/cli/parse.go#L36) is not nil-safe, so we should probably wrap the call in a `nil` check.

https://play.golang.org/p/T0OOtNMPzO

```golang
app := cli.NewApp()
app.Name = "buggy-cli"

app.ContextConfig = func(ctx context.Context) context.Context {
	return nil
}

app.Run(os.Args) // This will NPE
```

PR #77, opened 2017-09-07
Add safety check to ensure that context is non-nil

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/77)


joshdk:
This fixes #76

PR #78, opened 2017-10-04
tlsconfig: Explicitly disable client renegotiation

cbarcenas:
Explicitly disable TLS client renegotiation per internal guidelines.

### DIFF
--- a/tlsconfig/common.go
+++ b/tlsconfig/common.go
@@ -82,6 +82,7 @@ func configureTLSConfig(cfgs ...configurer) (*tls.Config, error) {
 		MinVersion:               tls.VersionTLS12,
 		PreferServerCipherSuites: true,
 		CipherSuites:             defaultCipherSuites,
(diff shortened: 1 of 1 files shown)


cbarcenas:
The intention here is not to completely prevent use of renegotiation. The change is intended to ensure renegotiation will always be disabled _by default_ (for the reasons listed below). As you note, renegotiation can still be explicitly enabled by consumers of `tlsconfig` and that's fine. The point of `tlsconfig` is to provide a secure baseline for TLS.

> Can you explain why this setting should never be used anywhere? tlsconfig is a general purpose TLS configuration library and I'm curious why it should impose this opinion.

Renegotiation has historically caused a number of security issues in TLS protocol versions and in some TLS implementations for newer versions due to the complexity it introduces in the TLS state machine.

There's some information in the [Go `tls` docs](https://golang.org/pkg/crypto/tls/#RenegotiationSupport) and I can also share internal TLS standards with you which mandate that it be disabled in all of our products.

PR #86, opened 2017-12-21
tlsconfig: add ability to set server's NextProtos

jdhenke:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/86)


nmiyake:
Thanks!

PR #116, opened 2018-10-03
Fix race condition in Each()

ashrayjain:
First commit adds a test that consistently panics due to the Each() implementation handling concurrent Unregister calls correctly.

Second commit adds a fix.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/116)

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -214,17 +214,19 @@ func (r *rootRegistry) Subregistry(prefix string, tags ...Tag) Registry {
 func (r *rootRegistry) Each(f MetricVisitor) {
 	// sort names so that iteration order is consistent
 	var sortedNames []string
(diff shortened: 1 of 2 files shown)


metrics/registry.go
```diff
@@ -238,10 +240,6 @@ func (r *rootRegistry) Each(f MetricVisitor) {
 			sort.Slice(tags, func(i, j int) bool {
 				return tags[i].String() < tags[j].String()
 			})
-		} else {
-			// if metric was not in idToMetricWithTags map, then it is a built-in metric. Trim the prefix since the
-			// registry lookup will automatically prepend the prefix.
-			metric = r.registry.Get(name)
```
ashrayjain: this part was redundant

PR #127, opened 2018-12-03
Add retry package

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/127)


retry/retry.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	for _, option := range opts {
+		option(&r.options)
+	}
+	r.Reset()
+	return r
+}
+
+const (
+	defaultMaxAttempts         = 0 // Infinite retries.
```
bmoylan: I do not like this default value (it blocks forever), but worried that changing it would be a break when we have internal consumers move to this import.

PR #128, opened 2018-12-03
Add bytesbuffers package

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/128)


bmoylan:
@nmiyake I added an example - do you think that is demonstrative enough?

nmiyake:
Yes, that is much more helpful -- thanks!

PR #134, opened 2018-12-19
Add rid constructors

k-simons:
- Add rid constructors for when you know the pieces of the identifier 
- Add additional tests (caught no incorrect functionality 👍  )

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/134)

### DIFF
--- a/rid/resource_identifier.go
+++ b/rid/resource_identifier.go
@@ -39,15 +39,15 @@ const (
 	Separator = "."
 )
 
-func MustOf(service, instance, resourceType, locator string) ResourceIdentifier {
-	resourceIdentifier, err := Of(service, instance, resourceType, locator)
+func MustNew(service, instance, resourceType, locator string) ResourceIdentifier {
+	resourceIdentifier, err := New(service, instance, resourceType, locator)
 	if err != nil {
 		panic(err)
 	}
 	return resourceIdentifier
 }
 
-func Of(service, instance, resourceType, locator string) (ResourceIdentifier, error) {
+func New(service, instance, resourceType, locator string) (ResourceIdentifier, error) {
 	resourceIdentifier := ResourceIdentifier{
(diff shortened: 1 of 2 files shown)


nmiyake:
___
*[rid/resource_identifier.go, line 38 at r2](https://reviewable.io/reviews/palantir/pkg/134#-LU6ayvM1uV4Tm4Sq5US:-LU6ayvM1uV4Tm4Sq5UT:b-3b27i5) ([raw file](https://github.com/palantir/pkg/blob/f31ccab4d620a7a2cf5c22f9444c1125fd43a3d5/rid/resource_identifier.go#L38)):*
> ```Go
> 
> const (
> 	RidClass  = "ri"
> ```

Is there a reason these constants need to be exported? Seems like they should be private.

Also, if exported, should be `RIDClass`.

nmiyake:
___
*[rid/resource_identifier.go, line 42 at r2](https://reviewable.io/reviews/palantir/pkg/134#-LU6aiQU6glGqccB5pZ-:-LU6aiQU6glGqccB5pZ0:b-ek2ove) ([raw file](https://github.com/palantir/pkg/blob/f31ccab4d620a7a2cf5c22f9444c1125fd43a3d5/rid/resource_identifier.go#L42)):*
> ```Go
> )
> 
> func MustNew(service, instance, resourceType, locator string) ResourceIdentifier {
> ```

Where is `MustNew` intended to be used? I would prefer not to include this as part of the actual API (if there's a client/tests that want to do this, they can easily implement this on their end).

nmiyake:
I see that this merged, but have some feedback that I think we should work through/address before cutting a release.

k-simons:
@nmiyake will open a follow up to discuss

k-simons:
https://github.com/palantir/pkg/pull/135

PR #143, opened 2019-01-25
metrics: copy tags list before serializing

bmoylan:
Fixes a bug introduced in #130 where the `tags` slice can be mutated by `sort.Sort`, affecting caller code unexpectedly.

This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/143)

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -385,7 +385,5 @@ func toMetricTagsID(name string, tags Tags) metricTagsID {
 
 // We copy the tag slice so that subsequent in-place mutations (sorting) do not affect the input slice.
 func copyTags(tags Tags) Tags {
-	tagsCopy := make(Tags, len(tags))
-	copy(tagsCopy, tags)
-	return tagsCopy
+	return append(tags[:0:0], tags...) // https://github.com/go101/go101/wiki/How-to-efficiently-clone-a-slice%3F
 }
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -305,31 +305,31 @@ func (r *rootRegistry) Unregister(name string, tags ...Tag) {
 }
 
 func (r *rootRegistry) Counter(name string, tags ...Tag) metrics.Counter {
-	return metrics.GetOrRegisterCounter(r.registerMetric(name, tags), r.registry)
+	return metrics.GetOrRegisterCounter(r.registerMetric(name, copyTags(tags)), r.registry)
 }
 
 func (r *rootRegistry) Gauge(name string, tags ...Tag) metrics.Gauge {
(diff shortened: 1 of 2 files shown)


bmoylan:
```
goos: darwin
goarch: amd64
pkg: github.com/palantir/pkg/metrics
BenchmarkRegisterMetric/1_tag-8                  5000000               374 ns/op             256 B/op          5 allocs/op
BenchmarkRegisterMetric/10_tag-8                 2000000               832 ns/op             848 B/op          5 allocs/op
BenchmarkRegisterMetric/100_tag-8                 100000             13932 ns/op            7568 B/op          5 allocs/op
PASS
ok      github.com/palantir/pkg/metrics 6.466s
```
The new allocation shows up as expected

ashrayjain:
@nmiyake just pointing out from the https://github.com/go101/go101/wiki/How-to-efficiently-clone-a-slice%3F

``` If the slice element type is string or non-basic types, then the efficiency differences between the two ways are small. For some cases, the append way might be even slower.)```

nmiyake:
___
*[metrics/registry.go, line 344 at r1](https://reviewable.io/reviews/palantir/pkg/143#-LX5chGb1tclsg-y3Og4:-LX5hY5YBsf4ZvlZNKfQ:b-eobyp1) ([raw file](https://github.com/palantir/pkg/blob/cd07bb8dae760b007dc6a5bf3e706729da13e116/metrics/registry.go#L344)):*

Haha I had it down here originally (https://github.com/palantir/pkg/commit/97c14c017979e403b2465aede67e7d35e2f61ca3) and @ashrayjain asked me to move it up so it would be more obvious to readers of the exported functions that the copy is happening for them.

Moved to the append style below but we should resolve this discussion of placement somehow (I'm relatively indifferent)
</blockquote></details>

I disagree -- feel pretty strongly that it should be here in `registerMetric`. That's where the common logic flows, so makes much more sense to centralize here.

For example, right now `Histogram` will copy the tags twice (because it delegates to `HistogramWithSample`, which also copies).

If we want clients to know that the slice will not be mutated, we can do that via function documentation. I think that code reading at that level is the wrong thing to optimize on (it's a bit more work, but someone reading the code can see that the only mutation happens after a copy).

ashrayjain:
@bmoylan @nmiyake i'm fine with keeping the copy in `registerMetric`.

nmiyake:
I was recommending the slice copy notation mainly to allow it to be a one-liner (wasn't really concerned on the efficiency front), but ultimately I'm fine with anything. Main point of feedback is that I think it should be inline in the one location it's used rather than creating a new function for it.

PR #152, opened 2019-03-18
Metrics: creating tags with fallback value

asanderson15:
We frequently want to create a tag with a known key, but potentially invalid value, in which case we want to fall back to a known value such as the string `unknown`. This allows us to do so without repeatedly rewriting the same fallback code.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/152)

### DIFF
--- a/metrics/tag.go
+++ b/metrics/tag.go
@@ -76,7 +76,7 @@ func MustNewTag(k, v string) Tag {
 	return t
 }
 
-// MustNewTagWithKnownKey returns the result of calling NewTag, but panics only if the key input is invalid.
+// MustNewTagWithFallbackValue returns the result of calling NewTag, but panics only if the key input is invalid.
 // If the value input is invalid, it will be replaced with the fallback value.
 func MustNewTagWithFallbackValue(k, v, fallback string) Tag {
 	tag, err := NewTag(k, v)
--- a/metrics/tag.go
+++ b/metrics/tag.go
@@ -76,6 +76,16 @@ func MustNewTag(k, v string) Tag {
 	return t
 }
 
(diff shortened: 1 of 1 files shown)


metrics/tag.go
```diff
@@ -76,6 +76,16 @@ func MustNewTag(k, v string) Tag {
 	return t
 }
 
+// MustNewTagWithFallbackValue returns the result of calling NewTag, but panics only if the key input is invalid.
+// If the value input is invalid, it will be replaced with the fallback value.
+func MustNewTagWithFallbackValue(k, v, fallback string) Tag {
```
bmoylan: - I would prefer a name like `NewTagWithFallbackValue` since `Must*` functions usually panic (e.g. the standard library's `regexp.MustCompile`)
- Let's mention in the comment that the fallback is passed through MustNewTag and will panic if invalid.

bmoylan: - I would prefer a name like NewTagWithFallbackValue since Must* functions usually panic (e.g. the standard library's regexp.MustCompile)
- Let's mention in the comment that the fallback is passed through MustNewTag and will panic if invalid.

bmoylan:
Will give @nmiyake a chance to take a look but we'll get a release out soon

nmiyake:
Sorry for dragging on this -- it's because this PR exists in the zone where I don't love adding it to the API, but understand the current pain/code duplication and can't say that it doesn't make sense to add out-right :)

At the end of the day, I don't see a major down-side to adding this, so OK with it on the whole. Left some feedback on preference to update the documentation a bit, but otherwise I'm good with this.

asanderson15:
No problem @nmiyake. That all makes sense. Agree with the comment and added the add'l docs.

PR #157, opened 2019-05-02
Add merge package for merging maps.

jamesross3:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/157)

### DIFF
--- a/merge/merge.go
+++ b/merge/merge.go
@@ -24,7 +24,7 @@ func Maps(dest, src interface{}) (interface{}, error) {
 // with the merge result for the return value. For map entries with the same key, the following rules apply:
 // 1. If the values have different types, the value from 'src' is used.
 // 2. If the src value is nil, the entry is absent from the resulting map.
-// 3. If the values are the same type and are structs, then src's value for each struct field is used, excepting map type fields, which are merged.
+// 3. If the values are the same type and are structs, then src's value for each struct field is used, excepting map type fields, which are recursively merged.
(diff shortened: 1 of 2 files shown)


jamesross3:
@bmoylan @k-simons FYSA

jamesross3:
@nmiyake thanks for pointing out mergo! This meets my use cases which means I don't think we need this change in pkg. Closing.

bmoylan:
Reopening this for @jamesross3 since the repo is currently restricted. From offline convo:
> mergo doesn't handle deep merging of `map[string]interface`s where the key is the same but the values for the key are different types

@jamesross3, can you add and point to a test case which mergo fails to handle?

nmiyake:
___
*[merge/merge.go, line 27 at r4](https://reviewable.io/reviews/palantir/pkg/157#-LesTDbK5OW0bxQ__4oi:-LesTDbK5OW0bxQ__4oj:bfr2ms8) ([raw file](https://github.com/palantir/pkg/blob/80973af93f6cb980e08fb721aa528294486a17b9/merge/merge.go#L27)):*
> ```Go
> // 1. If the values have different types, the value from 'src' is used.
> // 2. If the src value is nil, the entry is absent from the resulting map.
> // 3. If the values are the same type and are structs, the value of dest is persisted (this method is not intended to handle struct value merging).
> ```

This seems strange to me -- in every other case, the value in `src` is used. Why is it that, in the case of structs, the value from `dest` is used instead?

jamesross3:
test comment

PR #158, opened 2019-05-03
add IntParam type

nnchang:
basically a copy of the `StringParam` type, but with the int-parsing from `IntFlag`.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/158)


bmoylan:
Talked to nick offline, who says he no longer needs this

PR #174, opened 2019-10-08
Use microseconds as unit for Timer metrics

### LINKED ISSUE #173: metrics Timer should use microseconds
Currently, the `metrics.Registry` interface has a `Timer` function that creates a timer metric. The unit for this metric is currently nanoseconds, but our tooling generally uses microseconds for durations. For consistency, the timer implementation returned by `Timer` should record its units in microseconds.

Note that this change will alter the semantic output of the metrics registry.

nmiyake:
Fixes #173

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/174)


bmoylan:
:+1: 1.0 sounds good to me

PR #178, opened 2019-11-05
Add additional option to use fewer runtime metrics

andybradshaw:
Not super sure how I feel about this PR at the moment, but wanted to provide some ability for users to opt-out of getting some of the runtime stats that they might not care about.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/178)

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -35,6 +35,34 @@ var (
 		"go.runtime.NumCgoCall":            {},
 		"go.runtime.MemStats.PauseTotalNs": {},
 	}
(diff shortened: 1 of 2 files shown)


nmiyake:
I believe this is related to https://github.com/palantir/witchcraft-go-server/issues/125, and we have a PR open for that at https://github.com/palantir/witchcraft-go-server/pull/129.

If this is trying to address the same underlying issue, I think I would prefer to control this at the consumer level (in this example, in witchcraft-go-server) rather than at the producer level here.

andybradshaw:
Cool... think that's fair here.

PR #179, opened 2019-11-19
Transition to modules

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/179)


nmiyake:
Requesting feedback on this initial PR for module organization. Refer to internal communications for further details.

This PR should not break any existing consumers. It models each package as a separate module. The packages that depend on other packages in the repo use dep (they will be converted to modules once the modules here are tagged).

Each package has its own CircleCI job that runs "verify" with the current version of Go (with tests), and then has a dependent job that runs just tests using the previous version of Go (they are not all in parallel because it looks like CircleCI has an upper limit for number of jobs that is hit when every job here is its own job). Example CircleCI run at https://circleci.com/workflow-run/a7fe71c9-4ccf-4de5-b56e-6be49152f2ff.

This PR includes a simple go program that generates the CircleCI file in the current format for each module in the repository. The CircleCI job will fail if the diff is not empty. This is to make updating things like Go version in the future easy (and to properly handle case where top-level modules change). It is invoked directly since there is not a top-level godel project and because the operation is fairly simple (and this isn't a pattern we plan to repeat elsewhere).

.circleci/generate.go
```diff
@@ -0,0 +1,146 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"text/template"
+)
+
+const (
+	header = `checkout-path: &checkout-path
+  checkout-path: /go/src/github.com/palantir/pkg
```
bmoylan: Any value in proving modules are doing their thing by not using a GOPATH?

.circleci/generate.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func modules(parentDir string) ([]string, error) {
+	fis, err := ioutil.ReadDir(parentDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read directory: %v", err)
+	}
+	var dirNames []string
+	for _, fi := range fis {
+		if !fi.IsDir() || strings.HasPrefix(fi.Name(), ".") {
```
bmoylan: nit, probably doesn't matter in practice, but maybe we should verify there are .go files or go.mod so we don't include any directory that isn't a go module

Gopkg.lock
```diff
@@ -1,194 +0,0 @@
-# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
-
```
bmoylan: Any significant dependency version changes in the switch or not really? hard to tell from the diff

nmiyake:
I think it's a case-by-case basis, but my general thought is that:

* If the packages are a logical unit that should be versioned/released together, the repository should be a single module
* If the packages are truly independent and should not depend on each other, would recommend migrating via one of 2 approaches:
1. Same as here: convert current packages to v1 modules, and plan to promote to their own repo when they hit a v2. This provides the most backwards and forwards compatibility with GOPATH clients.
2. Create a new repository for the package when switching to modules
  * Repositories that convert to modules will need to know to depend on the new location
  * When existing consumers (that consume via the old monorepo) want to pick up new fixes, they will need to update import paths (treated as major version import)

The proper approach depends on how smooth you want upgrades for existing consumers of the package to be.

The option we have decided not to pursue is recommending multi-module repo for new/active development. Main reasoning is that it's hard for people to reason about and tag workflows are delicate as well. For full support, we would want to update autorelease-bot to support this as well, but that connotes endorsement to some degree so I think our current plan strikes the best balance. If the few other package repos want to follow this pattern, they can basically copy-paste/adopt the same approach we have here (since I believe n<5 I don't have major concerns with this).

If you want to talk through possible approaches for other existing repos let me know and would be happy to do that.

.circleci/generate.go
nmiyake: Yeah, when everything is a module we can check for `go.mod`, but right now a few of the projects are still non-module. Checking for `.go` would work for this repo, although in the general case the base package itself doesn't have to have a `.go` file (it could just have other packages within it). Will keep this for now given the one-off nature.

.circleci/generate.go
nmiyake: In Go 1.13 and above, if there's a `go.mod` present, module mode will always be used (even in GOPATH). We still need to keep the checkout path here in this commit since we have the non-module packages (which will be converted in next pass).

PR #182, opened 2019-12-02
Add top-level main.go file

nmiyake:
Allows the github.com/palantir/pkg module to be declared as a
module dependency, which can be used to resolve duplicate import
path issues.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/182)

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,0 +1,6 @@
+// +build module
+
+// This file exists only to smooth the transition for modules. Having this file allows other modules to declare a
+// dependency on the github.com/palantir/pkg module, which provides a mechanism for avoiding duplicate import path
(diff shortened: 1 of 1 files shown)


nmiyake:
This will be followed by tagging the repository as v1.0.1 and then updating the individual modules to import the `github.com/palantir/pkg` module at v1.0.1.

Doing this will make life easier for downstream consumers by ensuring that there are no import path conflicts due to another dependency pulling in a `github.com/palantir/pkg` pre-module dependency (it gets rid of the need for confusing "replace" directives).

This approach was suggested in https://github.com/golang/go/issues/35884#issuecomment-560547758, and I've verified that it works as expected by implementing the pattern in a similar repository at https://github.com/nmiyake/pkg.

PR #186, opened 2020-06-08
Change backoff semantics in case of initialBackoff

willdeuschle:
A not uncommon pattern is to use an initial backoff that is larger than the default max backoff. For example:

```
err = retry.Do(ctx, func() error {
	// do things
}, retry.WithMaxAttempts(5), retry.WithInitialBackoff(15*time.Second), retry.WithMultiplier(1))
```

I'd expect this to retry 5 times, every 15 seconds. However, because the default backoff is 2s, we end up retrying every two seconds.

The proposed change is to use the larger of max backoff or initial backoff in situations where the current backoff is larger than the max backoff.

Explicitly setting your initial backoff larger than your max backoff is nonsensical anyways, so I haven't been able to come up with an example where this change is detrimental.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/186)

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -124,6 +124,7 @@ func WithMaxAttempts(maxAttempts int) Option {
 // WithInitialBackoff sets initial backoff.
 //
 // If initial backoff option is not used, then default value of 50 milliseconds is used.
+// If initial backoff is larger than max backoff, it takes precedence.
 func WithInitialBackoff(initialBackoff time.Duration) Option {
 	return func(o *options) {
 		o.initialBackoff = initialBackoff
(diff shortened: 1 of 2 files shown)


retry/retry.go
```diff
@@ -247,7 +247,11 @@ func (r *retrier) Next() bool {
 func (r retrier) retryIn() time.Duration {
 	backoff := float64(r.options.initialBackoff) * math.Pow(r.options.multiplier, float64(r.currentAttempt))
 	if r.options.maxBackoff != 0 && backoff > float64(r.options.maxBackoff) {
-		backoff = float64(r.options.maxBackoff)
+		if r.options.initialBackoff > r.options.maxBackoff {
```
bmoylan, line 250: Would it be better to set `maxBackoff = max(maxBackoff, initialBackoff)` in `Start()` after applying the params so we don't do this check every time, given that the logic would be the same? I might consider this semantic more "construction-time" than "runtime"

bmoylan, line 250: Can you include a comment on the InitialBackoff param explaining that if it's larger than MaxBackoff, it will take precedent?

willdeuschle: I considered that and ultimately chose not to do so because it mucks with what is currently a very clean/understandable `Start` with initialization and then application of option. I'm happy to put it there instead

willdeuschle: will do

PR #189, opened 2020-07-28
stop overriding max backoff when none is set

bryanhanner:
When adding functionality to override the max backoff, we need to special case when no backoff is set.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/189)

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -115,6 +115,8 @@ type Option func(r *options)
 // Max attempts of 0 indicates no limit.
 //
 // If max attempts option is not used, then default value of 0 is used.
+// If initial backoff is larger than max backoff and the max backoff is nonzero, the initial backoff will be
+// used as the max.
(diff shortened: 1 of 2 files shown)


retry/retry.go
```diff
@@ -115,6 +115,8 @@ type Option func(r *options)
 // Max attempts of 0 indicates no limit.
 //
 // If max attempts option is not used, then default value of 0 is used.
+// If initial backoff is larger than max backoff and the max backoff is nonzero, the initial backoff will be
```
ashrayjain, line 118: this is the wrong place for this godoc?

bryanhanner: True, people should read the initialBackoff docs if they use it, so need for a dup.

bmoylan:
// linking #186 and cc @willdeuschle

PR #192, opened 2020-09-10
AllowRoot setting to control if a cli is allowed to run as root

montaguethomas:
Adds new `AllowRoot` boolean setting to control if an implementation of cli App is allowed to run as UID 0 (root).
Defaults to `false`.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -25,6 +25,7 @@ type App struct {
 	OnExit         OnExit
 	ContextConfig  func(Context, context.Context) context.Context
 	ContextOptions []ContextOption
+	AllowRoot      bool
 }
 
 type Option func(*App)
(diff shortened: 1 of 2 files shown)


bmoylan (changes requested):
Defaulting to false would be a breaking change for anything that _wants_ to run as root and gets upgraded without knowledge of the breaking change. If we go this route it should maintain the same default behavior and opt-in to the stricter version.

That said, I'm not that keen on this solution to what was ultimately a bug in the infrastructure that should be fixed. The internal spawning daemon that caused the issue this is supposed to address already has this check in application code (added in April), so I'm not sure in what scenario this would help going forward.

gregakinman:
@bmoylan I could see value in this as serving as a more general upstream implementation of the solution that was merged to the internal spawning daemon (since it uses this library so could just change its current implementation to using this feature)

PR #194, opened 2021-01-12
Fix ClientKeyPair param in tlsconfig so clients refresh by default

Neverblade:
Golang TLS supports `GetClientCertificate` in tls config, which calls a function when a certificate is requested instead of serving a static certificate. We change the `ClientKeyPair` param to use this field instead of setting a static certificate. `ClientKeyPairFiles` remains static. This enables users to refresh their certificates so they don't expire.

Comments have been updated to clarify the differences between params.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/194)

### DIFF
--- a/tlsconfig/tlsconfig_client.go
+++ b/tlsconfig/tlsconfig_client.go
@@ -32,11 +32,11 @@ func (p clientParam) configureClient(cfg *tls.Config) error {
 	return p(cfg)
 }
 
-// ClientKeyPairFiles configures the client with the key pair that it should present to servers when communicating using
+// ClientKeyPairFiles configures the client with a static key pair for it to present to servers when communicating using
 // TLS with client authentication (2-way SSL). If neither ClientKeyPairFiles nor ClientKeyPair are provided, the client
 // will not present a certificate.
 func ClientKeyPairFiles(certFile, keyFile string) ClientParam {
-	return ClientKeyPair(TLSCertFromFiles(certFile, keyFile))
+	return clientParam(certificatesParam(TLSCertFromFiles(certFile, keyFile)))
 }
 
 // ClientKeyPair configures the client to call the provided TLSCertProvider whenever a key pair is requested when
--- a/tlsconfig/tlsconfig_client.go
+++ b/tlsconfig/tlsconfig_client.go
@@ -39,9 +39,9 @@ func ClientKeyPairFiles(certFile, keyFile string) ClientParam {
 	return ClientKeyPair(TLSCertFromFiles(certFile, keyFile))
 }
 
-// ClientKeyPair configures the client with the static key pair that it should present to servers when communicating
-// using TLS with client authentication (2-way SSL). If neither ClientKeyPairFiles, ClientKeyPair, nor
-// ClientKeyPairProvider are provided, the client will not present a certificate.
+// ClientKeyPair configures the client with a static key pair sourced from the TLSCertProvider that it should present to
+// servers when communicating using TLS with client authentication (2-way SSL). If neither ClientKeyPairFiles,
+// ClientKeyPair, nor ClientKeyPairProvider are provided, the client will not present a certificate.
 func ClientKeyPair(certProvider TLSCertProvider) ClientParam {
 	return clientParam(certificatesParam(certProvider))
 }
--- a/tlsconfig/tlsconfig_client.go
+++ b/tlsconfig/tlsconfig_client.go
@@ -33,16 +33,24 @@ func (p clientParam) configureClient(cfg *tls.Config) error {
 }
 
 // ClientKeyPairFiles configures the client with the key pair that it should present to servers when communicating using
-// TLS with client authentication (2-way SSL). If this parameter is not provided, the client will not present a
-// certificate.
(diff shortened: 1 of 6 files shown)


tlsconfig/tlsconfig_client.go
```diff
@@ -33,16 +33,24 @@ func (p clientParam) configureClient(cfg *tls.Config) error {
...
 func ClientKeyPairFiles(certFile, keyFile string) ClientParam {
 	return ClientKeyPair(TLSCertFromFiles(certFile, keyFile))
 }
 
-// ClientKeyPair configures the client with the key pair that it should present to servers when communicating using TLS
-// with client authentication (2-way SSL). If this parameter is not provided, the client will not present a certificate.
+// ClientKeyPair configures the client with a static key pair sourced from the TLSCertProvider that it should present to
+// servers when communicating using TLS with client authentication (2-way SSL). If neither ClientKeyPairFiles,
+// ClientKeyPair, nor ClientKeyPairProvider are provided, the client will not present a certificate.
 func ClientKeyPair(certProvider TLSCertProvider) ClientParam {
```
bmoylan, line 45: Is there a reason to keep the static variant around? Can we just update the logic of this parameter to use `GetCertificate` and then everyone gets it for free?

Are there cases where the supplier function is expensive so we don't want to call it more than we have to?

jhowarth-pal: I don't know of any examples, but it's easy to imagine someone could use a non-cached vault based provider or even load from disk. @jlin one idea could be to make the refreshing version the default but also provide a static version? No strong opinions here though

bmoylan: Discussed on Slack, and since we can not find any examples of expensive cert providers (rather than returning from memory/cache) we are going to treat this change as a bugfix to make `ClientKeyPair` use `GetCertificate` and not add a separate param. cc @Neverblade

tlsconfig/tlsconfig_server.go
```diff
@@ -26,7 +26,7 @@ func NewServerConfig(tlsCertProvider TLSCertProvider, params ...ServerParam) (*t
 	if tlsCertProvider == nil {
 		return nil, fmt.Errorf("tlsCertProvider provided to NewServerConfig was nil")
 	}
-	configurers := []configurer{authKeyPairParam(tlsCertProvider)}
+	configurers := []configurer{certificatesParam(tlsCertProvider)}
```
Neverblade, line 29: Wanted to make a parallel change for servers too because Golang TLS conf has a similar `GetCertificate` func, however ran into an issue that stems from this logic that determines if `GetCertificate` gets called:
```
	if c.GetCertificate != nil &&
		(len(c.Certificates) == 0 || len(clientHello.ServerName) > 0) {
		cert, err := c.GetCertificate(clientHello)
		if cert != nil || err != nil {
			return cert, err
		}
	}
```

As an example, [tlsconfig_test](https://github.com/palantir/pkg/blob/master/tlsconfig/tlsconfig_test.go) doesn't set a servername, and the `StartTLS()` func adds a local host cert when it detects that the certificates list is empty. This means the tests fails as the local host cert isn't recognized by the client. It's possible that this edge case could happen in the wild as well

Don't want to block this PR so I'm scoping this PR to client behavior specifically, but worth considering the server side as a FLUP

tlsconfig/tlsconfig_client.go
Neverblade: Done

bmoylan (approved):
> Reviewed 1 of 4 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/pkg/194)* status: all files reviewed, 1 unresolved discussion (waiting on @Neverblade)

___
*[tlsconfig/tlsconfig_server.go, line 29 at r2](https://reviewable.io/reviews/palantir/pkg/194#-MQsP94J5BeB-78YHT8q-r2-29:-MQxa4dyA_8S-PecfTCb:b-p07id9) ([raw file](https://github.com/palantir/pkg/blob/e5203522dc21233587fe14173bd19429356ac92a/tlsconfig/tlsconfig_server.go#L29)):*
<details><summary><i>Previously, Neverblade (James Lin) wrote…</i></summary><blockquote>

Wanted to make a parallel change for servers too because Golang TLS conf has a similar `GetCertificate` func, however ran into an issue that stems from this logic that determines if `GetCertificate` gets called:
```
	if c.GetCertificate != nil &&
		(len(c.Certificates) == 0 || len(clientHello.ServerName) > 0) {
		cert, err := c.GetCertificate(clientHello)
		if cert != nil || err != nil {
			return cert, err
		}
	}
```

As an example, [tlsconfig_test](https://github.com/palantir/pkg/blob/master/tlsconfig/tlsconfig_test.go) doesn't set a servername, and the `StartTLS()` func adds a local host cert when it detects that the certificates list is empty. This means the tests fails as the local host cert isn't recognized by the client. It's possible that this edge case could happen in the wild as well

Don't want to block this PR so I'm scoping this PR to client behavior specifically, but worth considering the server side as a FLUP
</blockquote></details>

Can you file an issue with these details so we don't lose the context?

PR #196, opened 2021-01-24
Add refreshable package, ported from witchcraft-go-server

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/196)


refreshable/main.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2020 Palantir Technologies. All rights reserved.
```
tabboud, line 1: I believe you need to re-run `license` since this is still `2020`

refreshable/refreshable_types.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+func NewIntPtr(in Refreshable) IntPtr {
+	return refreshableTyped{
+		Refreshable: in,
+	}
+}
+
+var _ Bool = (*refreshableTyped)(nil)
```
tabboud, line 111: nit: let's use a `var ()` block instead

tabboud (commented):
It looks like this is not a pure 1:1 port, but a port + some additional types added. We could separate the port and wiring from the added types, but this seems fine to me, given the underlying logic appears to be unchanged. Overall the change looks good to me.

Did you think about create a new repo to house this instead of pkg? I vaguely remember @nmiyake mentioning that we should consider single repos for newly added features due to weirdness with nested modules, but I will defer to him.

bmoylan:
Thanks @tabboud for clarity I reverted most of the new types and moved them to #198. As far as repository location, my mental model for `pkg` packages is things that are self-contained (few if any dependencies), relatively small, and would benefit from being the "one true global package" for X (e.g. in the case of shared type packages `bearertoken`, `safelong`, `binary`, etc). This is also my justification for #199.

As far as the multi-module setup, I agree its a little funky but we have it solved here where adding a new package is much simpler than adding a new repository through the oss-approvals process for projects that are only a couple files.

tabboud (approved):
Ok that makes sense to me. I generally agree this feels like the right place to live, especially given we have the multi-modules infra setup.

PR #199, opened 2021-01-29
Add yamlpatch package for applying JSONPatch to YAML gracefully

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/199)


yamlpatch/patch.go
```diff
@@ -0,0 +1,255 @@
...
+	"encoding/json"
+	"reflect"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+var (
+	// IndentSpaces configures the yaml encoder used to marshal patched objects.
+	IndentSpaces = 2
```
nmiyake, line 20: Is this variable intentionally exported so that consumers can modify it? If so, does it make sense to be global for the package, or to have this be something that is configured per operation?

nmiyake (commented):
Thanks for submitting this!

Question about the exported `IndentSpaces` variable, but otherwise this looks good!

nmiyake:
@bmoylan tagging because it's been a while since this PR was opened but finally got a chance to review.

yamlpatch/patch.go
bmoylan: Yeah, admittedly this was me being a little lazy with how to configure this. I've just pushed a commit which unexports this variable in favor of an additional `ApplyIndent` entrypoint which allows specifying the indent length.

nmiyake: Thanks!

Mostly on-board, but just wanted to discuss how to handle parameters in general before merging this since, while playing around with this locally, there was another thing I wanted to configure.

Right now, the value of the Patch is converted to a YAML node by using `json.Marshal`. However, this isn't ideal for Go structs that define YAML marshalling -- behavior like `omitempty`, custom field names and field ordering are all not respected. Since this is likely to be used by Go code that operates on YAML structs, I think that wanting to marshal while respecting this is a common use case. Although performing a YAML struct -> YAML bytes -> JSON struct translation on the caller side solves a lot of the issues, in this case you lose field ordering.

This problem can be solved by using `yaml.Marshal` rather than `json.Marshal` in `valueToYAMLNode`, and I think that this is something that is reasonable to be able to configure (either as "YAML mode" or more generally by allowing callers to provide an `interface{} -> []byte` function).

Given that we now have scenarios where we may want to configure indentation and marshal mode and there may be more in the future, wondering if we should go an OO route where we configure a marshaler and have `Apply` be a function on the marshaler, or have `Apply` accept `Param` objects and define these configurable things that way.

yamlpatch/patch.go
```diff
@@ -0,0 +1,262 @@
...
+			err = patchTest(node, op.Path, op.Value)
+		default:
+			err = errors.Errorf("unexpected op")
+		}
+		if err != nil {
+			return nil, errors.Wrapf(err, "op %s %s", op.Type, op.Path.String())
+		}
+	}
+	buf := bytes.Buffer{}
+	enc := yaml.NewEncoder(&buf)
```
nmiyake, line 64: Documentation for `yaml.NewEncoder` states that it should be closed after use to flush all data, so would add deferred close.

yamlpatch/patch.go
```diff
@@ -0,0 +1,264 @@
...
+// Apply the patch to a yaml document provided in originalBytes and return the updated content.
+// A best effort is made to minimize changes outside the patched paths but some whitespace changes are unavoidable.
+func Apply(originalBytes []byte, patch Patch) ([]byte, error) {
+	return ApplyIndent(originalBytes, patch, defaultIndentSpaces)
+}
+
+// ApplyIndent the patch to a yaml document provided in originalBytes and return the updated content.
+// indentSpaces configures the indentation used to marshal patched objects.
+// A best effort is made to minimize changes outside the patched paths but some whitespace changes are unavoidable.
+func ApplyIndent(originalBytes []byte, patch Patch, indentSpaces int) ([]byte, error) {
```
nmiyake, line 31: Thought about this a bit more, and propose the following path forward:

* As part of this initial PR, let's just remove `ApplyIndent` and only have the `Apply` function. This means that indentation level can't be configured as part of the initial PR, but think that should be fine.
* I will do this as a follow-up PR, but I propose that we follow the general pattern of things like `json.Marshal` here and effectively have a `PatchApplier` struct that can be created and configured (whether it be via params or functions) to apply patches. The applier can be configured to do things like set indentation level, control whether values are marshaled as JSON or YAML, etc.
* Implementation of `ApplyIndent` will be updated to effectively use the default `PatchApplier` struct/settings

nmiyake (changes requested):

nmiyake (approved):
Now that the marshal is done via YAML rather than JSON, I can't think of another specific configuration point that is needed at this time. Given that the builtin `json.Marshal` also has an `Indent` variant, I think it's reasonable to proceed here -- if we want to add a new `PatchApplier` struct or use the Param paradigm, it will still be simple enough to introduce that while still supporting the current API, so think we can move forward. Thanks!

nmiyake:
This PR actually merged -- however, GitHub was having issues at the time so looks like the merge happened but the PR itself was not closed. Closing this PR to indicate that.

PR #201, opened 2021-02-03
refreshable: Add Map and Subscribe methods to builtins

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/201)

### DIFF
--- a/refreshable/refreshable_types.go
+++ b/refreshable/refreshable_types.go
@@ -68,6 +68,8 @@ type Duration interface {
 type DurationPtr interface {
 	Refreshable
 	CurrentDurationPtr() *time.Duration
+	MapDurationPtr(func(time.Duration) interface{}) Refreshable
+	SubscribeToDurationPtr(func(time.Duration)) (unsubscribe func())
 }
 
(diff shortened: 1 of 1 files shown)


refreshable/refreshable_types.go
```diff
@@ -11,42 +11,58 @@ import (
...
 	CurrentBoolPtr() *bool
+	MapBoolPtr(func(*bool) interface{}) Refreshable
+	SubscribeToBoolPtr(func(*bool)) (unsubscribe func())
 }
 
 // Duration is a Refreshable that can return the current time.Duration.
 type Duration interface {
 	Refreshable
 	CurrentDuration() time.Duration
+	MapDuration(func(time.Duration) interface{}) Refreshable
```
tabboud, line 64: We should also add the `MapDurationPtr` and `SubscribeToDurationPtr` below.

tabboud (changes requested):
This looks great, thanks for adding it in!
Just one comment on the missing methods

tabboud (approved):

PR #203, opened 2021-03-18
Make tls.InsecureSkipVerify configurable.

jamesross3:
It is currently not possible to use the tlsconfig package without requiring secure TLS. Surfacing this option enables rare but necessary off-roading.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/203)

### DIFF
--- a/tlsconfig/tlsconfig_client.go
+++ b/tlsconfig/tlsconfig_client.go
@@ -73,7 +73,10 @@ func ClientCipherSuites(cipherSuites ...uint16) ClientParam {
 	return clientParam(cipherSuitesParam(cipherSuites...))
 }
 
-func InsecureSkipVerify() ClientParam {
+// ClientInsecureSkipVerify sets the InsecureSkipVerify field of tls Config to true. The default value for this field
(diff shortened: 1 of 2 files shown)


tlsconfig/tlsconfig_client.go
```diff
@@ -72,3 +72,10 @@ func ClientRootCAs(certPoolProvider CertPoolProvider) ClientParam {
 func ClientCipherSuites(cipherSuites ...uint16) ClientParam {
 	return clientParam(cipherSuitesParam(cipherSuites...))
 }
+
+func InsecureSkipVerify() ClientParam {
```
tabboud, line 76: Can you update this to follow the `Client` naming convention -> `ClientInsecureSkipVerify()`.
Additionally can you add a doc comment that includes the default value if not set.

tabboud (changes requested):

tlsconfig/tlsconfig_client.go
jamesross3: Done

tabboud (approved):

PR #206, opened 2021-04-08
refreshable: Add MapValidatingRefreshable

bmoylan:
This allows for a transformation function to return an error.

### DIFF
--- a/refreshable/refreshable_validating.go
+++ b/refreshable/refreshable_validating.go
@@ -11,26 +11,12 @@ import (
 
 type ValidatingRefreshable struct {
 	Refreshable
-
-	validatedRefreshable Refreshable
-	lastValidateErr      *atomic.Value
+	lastValidateErr *atomic.Value
 }
 
 // this is needed to be able to store the absence of an error in an atomic.Value
 type errorWrapper struct {
 	err error
-}
-
-func (v *ValidatingRefreshable) Current() interface{} {
-	return v.validatedRefreshable.Current()
-}
-
-func (v *ValidatingRefreshable) Subscribe(consumer func(interface{})) (unsubscribe func()) {
-	return v.validatedRefreshable.Subscribe(consumer)
-}
-
-func (v *ValidatingRefreshable) Map(mapFn func(interface{}) interface{}) Refreshable {
-	return v.validatedRefreshable.Map(mapFn)
 }
 
 func (v *ValidatingRefreshable) LastValidateErr() error {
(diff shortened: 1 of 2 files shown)


refreshable/refreshable_validating.go
```diff
@@ -41,6 +27,22 @@ func (v *ValidatingRefreshable) LastValidateErr() error {
...
+	mappingFn := func(i interface{}) (interface{}, error) {
+		if err := validatingFn(i); err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
+	return newValidatingRefreshable(origRefreshable, mappingFn, false)
+}
+
+// MapValidatingRefreshable is similar to NewValidatingRefreshable but allows for the function to return a mapping/mutation
```
tabboud, line 39: Can you also mention that the returned value is what is stored as the current value of the refreshable, rather than the current value of the original refreshable. It's a subtle nuance but seems important to call out for the docs.

refreshable/refreshable_validating.go
```diff
@@ -41,6 +27,22 @@ func (v *ValidatingRefreshable) LastValidateErr() error {
...
+// of the input object in addition to returning an error. It must always return a value of same type.
```
tabboud, line 40: The last sentence is not actually true and we don't enforce it. For example in the test you added it goes from `string -> *url.URL`

refreshable/refreshable_validating.go
```diff
@@ -41,6 +27,22 @@ func (v *ValidatingRefreshable) LastValidateErr() error {
...
+func MapValidatingRefreshable(origRefreshable Refreshable, mappingFn func(interface{}) (interface{}, error)) (*ValidatingRefreshable, error) {
```
tabboud, line 41: nit: new constructor -> `NewMapValidatingRefreshable`?

refreshable/refreshable_validating.go
```diff
@@ -41,6 +27,22 @@ func (v *ValidatingRefreshable) LastValidateErr() error {
...
+	return newValidatingRefreshable(origRefreshable, mappingFn, true)
+}
+
+func newValidatingRefreshable(origRefreshable Refreshable, validatingFn func(interface{}) (interface{}, error), storeVal bool) (*ValidatingRefreshable, error) {
```
tabboud, line 45: nit: `s/storeVal/storeMappedVal` to be more explicit about what we're storing?

tabboud (changes requested):

refreshable/refreshable_validating.go
bmoylan: Updated, lmk what you think

refreshable/refreshable_validating.go
bmoylan: Yeah, this is supposed to mean that the returned `interface{}` has to always be the same type so the DefaultRefreshable will accept it. I tried to tweak the language, lmk if it makes sense

refreshable/refreshable_validating.go
bmoylan: Done.

refreshable/refreshable_validating.go
bmoylan: Done.

refreshable/refreshable_validating.go
tabboud: ahh got it, that makes more sense. So basically the mapping function must be consistent with the returned type. For sake of example, this would be invalid as the type is changed if the default original is updated to anything but 1.
```go
r := refreshable.NewDefaultRefreshable(1)
vr, err := refreshable.MapValidatingRefreshable(r, func(i interface{}) (interface{}, error) {
		switch v := i.(int); v {
		case 1:
			return "1", nil
		default:
			return 2, nil
		}
	})
```

tabboud (approved):

PR #208, opened 2021-04-12
refreshable: fix: update non-mapping validating refreshable

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/208)


tabboud (approved):
good catch 👍

PR #218, opened 2021-08-16
improvement: Performance improvements to NewTag builder.

dtrejod:
## Before this PR
In go 1.10 `strings.Builder` [type](https://pkg.go.dev/strings#Builder) was introduced as an effective way to build strings. 

## After this PR
Refactor the NewTag builder to use the string Builder type and optimize iterations over the key/value pairs.
==COMMIT_MSG==
Performance improvements to NewTag builder.
==COMMIT_MSG==

**Benchmarks**
```
name                 old time/op    new time/op    delta
NewTag/tagLen:2-8       156ns ± 1%     127ns ± 0%   ~     (p=0.333 n=2+2)
NewTag/tagLen:10-8      694ns ± 0%     203ns ± 3%   ~     (p=0.333 n=2+2)
NewTag/tagLen:100-8    7.90µs ± 0%    0.91µs ± 4%   ~     (p=0.333 n=2+2)
NewTag/tagLen:199-8    16.8µs ± 0%     1.7µs ± 1%   ~     (p=0.333 n=2+2)

name                 old alloc/op   new alloc/op   delta
NewTag/tagLen:2-8       12.0B ± 0%      5.0B ± 0%   ~     (p=0.333 n=2+2)
NewTag/tagLen:10-8      96.0B ± 0%     24.0B ± 0%   ~     (p=0.333 n=2+2)
NewTag/tagLen:100-8    3.54kB ± 0%    0.24kB ± 0%   ~     (p=0.333 n=2+2)
NewTag/tagLen:199-8    12.1kB ± 0%     0.4kB ± 0%   ~     (p=0.333 n=2+2)

name                 old allocs/op  new allocs/op  delta
NewTag/tagLen:2-8        3.00 ± 0%      3.00 ± 0%   ~     (all equal)
NewTag/tagLen:10-8       19.0 ± 0%       3.0 ± 0%   ~     (p=0.333 n=2+2)
NewTag/tagLen:100-8       199 ± 0%         3 ± 0%   ~     (p=0.333 n=2+2)
NewTag/tagLen:199-8       395 ± 0%         3 ± 0%   ~     (p=0.333 n=2+2
```

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/218)


metrics/tag.go
```diff
@@ -151,20 +146,28 @@ func NewTags(t map[string]string) (Tags, error) {
 	return tags, nil
 }
 
-var validChars = make(map[rune]struct{})
+var validKeyChars = make(map[rune]struct{})
```
bmoylan, line 149: It seems like you can make this faster by using arrays and avoid the map access -- on my laptop this removed around 20% from the benchmark: https://github.com/palantir/pkg/compare/dt/new-tag-optomize...bm/dt/new-tag-optomize

metrics/tag.go
```diff
@@ -173,14 +176,15 @@ func init() {
 // unicode. Tags will be converted to lowercase."
 //
 // Note that this function does not impose the length restriction described above.
-func normalizeTag(in string) string {
+func normalizeTag(in string, validChars map[rune]struct{}) string {
+	var builder strings.Builder
+	builder.Grow(len(in))
 	in = strings.ToLower(in)
```
bmoylan, line 182: if we do the lower-casing rune-by-rune can we avoid allocating another string?

metrics/benchmark_test.go
```diff
@@ -10,19 +10,39 @@ import (
...
+	b.Run("100 tag", func(b *testing.B) {
+		doTagBench(b, 100)
+	})
+}
+
+func doTagBench(b *testing.B, n int) {
+	var tags Tags
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		tags = append(tags, MustNewTag(fmt.Sprintf("key%d", i), fmt.Sprintf("val%d", i)))
```
bmoylan, line 29: A lot of the time in the benchmark is going into growing this array. Just assigning the result to `_` will run the function with less happening that you aren't trying to observe
```suggestion
		n := strconv.Itoa(i)
		_ = MustNewTag("key"+n, "val"+n)
```

metrics/tag.go
bmoylan: Stole this from the ToLower impl, but I think you can just add it as an else statement where we check for valid chars:
```go
		} else if 'A' <= r && r <= 'Z' {
			r += 'a' - 'A'
```

metrics/benchmark_test.go
```diff
@@ -10,19 +10,39 @@ import (
...
+	})
+	b.Run("10 tag", func(b *testing.B) {
+		doTagBench(b, 10)
+	})
+	b.Run("100 tag", func(b *testing.B) {
+		doTagBench(b, 100)
+	})
+}
+
+func doTagBench(b *testing.B, n int) {
```
bmoylan, line 25: this `n` argument is getting ignored, is there supposed to be an inner loop?

bmoylan: actually its probably more interesting to test different string lengths to see how that affects things (is a 10-char string 10x slower than a 1-char string?

dtrejod (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/pkg/218)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @bmoylan)

___
*[metrics/benchmark_test.go, line 25 at r1](https://reviewable.io/reviews/palantir/pkg/218#-MhFh2dKYC8zw-UyN4Rl-r1-25:-MhGPrBm2ewHYt9HmU5p:bxjm8qh) ([raw file](https://github.com/palantir/pkg/blob/b0143b1becba73efc6a96ef62bb168d8a14b5ab7/metrics/benchmark_test.go#L25)):*

actually its probably more interesting to test different string lengths to see how that affects things (is a 10-char string 10x slower than a 1-char string?
</blockquote></details>

Doh you are right. You are right, testing the number of tags isn't as useful as testing the length of the tag strings. I updated the benchmark to reflect that. 
___
*[metrics/benchmark_test.go, line 29 at r1](https://reviewable.io/reviews/palantir/pkg/218#-MhFh2dKYC8zw-UyN4Rl-r1-29:-MhGQ13MBeuWs25wn2V0:b-w5ckpu) ([raw file](https://github.com/palantir/pkg/blob/b0143b1becba73efc6a96ef62bb168d8a14b5ab7/metrics/benchmark_test.go#L29)):*

A lot of the time in the benchmark is going into growing this array. Just assigning the result to `_` will run the function with less happening that you aren't trying to observe
```suggestion
		n := strconv.Itoa(i)
		_ = MustNewTag("key"+n, "val"+n)
```
</blockquote></details>

Not relevant anymore.
___
*[metrics/tag.go, line 149 at r1](https://reviewable.io/reviews/palantir/pkg/218#-MhFh2dKYC8zw-UyN4Rm-r1-149:-MhGQ6q6-Kqyxc0zRnWw:b-mqpcqu) ([raw file](https://github.com/palantir/pkg/blob/b0143b1becba73efc6a96ef62bb168d8a14b5ab7/metrics/tag.go#L149)):*

It seems like you can make this faster by using arrays and avoid the map access -- on my laptop this removed around 20% from the benchmark: https://github.com/palantir/pkg/compare/dt/new-tag-optomize...bm/dt/new-tag-optomize
</blockquote></details>

Nice! Surprised the map access had that much of an impact here. These changes 
___
*[metrics/tag.go, line 182 at r1](https://reviewable.io/reviews/palantir/pkg/218#-MhFh2dKYC8zw-UyN4Rm-r1-182:-MhGYQNG4ySx8dYYkr2a:bqmudqt) ([raw file](https://github.com/palantir/pkg/blob/b0143b1becba73efc6a96ef62bb168d8a14b5ab7/metrics/tag.go#L182)):*

Stole this from the ToLower impl, but I think you can just add it as an else statement where we check for valid chars:
```go
		} else if 'A' <= r && r <= 'Z' {
			r += 'a' - 'A'
```

</blockquote></details>

Yeah we can `unicode.ToLower` here to optimize and avoid the extra string.

bmoylan (approved):
:lgtm: thanks!

> Reviewed 1 of 2 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/palantir/pkg/218#-:-MhH5GtPE3VmaKka7wB-:brfqlux)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dtrejod)

PR #229, opened 2021-12-14
Upgrade Go versions to 1.17.5/1.16.12

tabboud:
This change only updates the Go version used in CircleCI and does not touch any of the go directives in the various `go.mod` files.

This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/229)

### DIFF
--- a/.circleci/generate.go
+++ b/.circleci/generate.go
@@ -1,7 +1,7 @@
 // +build generate
 
 // This program prints the CircleCI configuration for the "pkg" repository. Standard way to run it is to run
-// "go run generate.go".
+// "go run generate.go {{parentDir}} > config.yml".
 package main
 
 import (
--- a/.circleci/generate.go
+++ b/.circleci/generate.go
@@ -88,7 +88,7 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	configYML, err := createConfigYML(mods, "1.15.7", "1.14.14")
+	configYML, err := createConfigYML(mods, "1.17.5", "1.16.12")
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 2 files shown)


.circleci/generate.go
```diff
@@ -88,7 +88,7 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	configYML, err := createConfigYML(mods, "1.15.7", "1.14.14")
+	configYML, err := createConfigYML(mods, "1.17.5", "1.16.12")
```
nmiyake, line 91: We should probably have an excavator that performs this bump for us

bmoylan (approved):
Maybe we should update the go.mod directives if we aren't going to test older versions? You all are more on top of this than I am, but I think we've gotten almost everything on at least 1.16

nmiyake (approved):
Thanks! As noted in the review, would be nice if we had an excavator that handled this bump for us, but makes sense in the meantime.

nmiyake:
Also +1 to Brad's sentiment that, ideally, we should update the `go.mod` directives to be the lesser of the 2 versions as well. Ideal would probably be to upgrade the `go generate` task to handle that as well.

tabboud:
Thanks for the quick review! I will update the generate logic in this PR to also bump the go directives to the minimum version and will re-tag you both for review after.

Nick, that was my thought as well! I added a note to create this excavator and will follow up when I have some more free time.

tabboud:
Looks like the go directive in all sub-modules are already updated to `go 1.16` via the "Manage go version" excavator (https://github.com/palantir/pkg/pull/217), so there is no additional change needed.
I wrote the directive update logic, but will open a separate PR to merge that in considering there is no change here.

PR #230, opened 2021-12-14
Update all go.mod Go directives

tabboud:
This change is a flup from #229 to update all Go directives for every module. The Go versions used in CircleCI are managed in the [.circleci/generate.go](https://github.com/palantir/pkg/blob/master/.circleci/generate.go) file, but the go directives for all modules were not updated which meant the versions could fall behind what is used in Circle.

One thing to consider is that this repo is currently taking managed Go version updates via excavator (#217) however the CircleCI config does not leverage this. The excavator check also updates all of the go directives, so we may want to disable this repo from the managed Go version check in favor of using this generate script, at least until we converge on a single approach for multi-module management.

This PR also updates the usage of the command to use flags instead of just arguments so that the module updates are conditionally run.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/230)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       - checkout
       - run: go version
       - run: go run .circleci/generate.go -repoRoot .
-      - run: diff  <(cat .circleci/config.yml) <(go run .circleci/generate.go .)
+      - run: diff  <(cat .circleci/config.yml) <(go run .circleci/generate.go -repoRoot .)
   circle-all:
     docker:
       - image: "golang:1.17.5"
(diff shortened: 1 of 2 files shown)


nmiyake:
Closing this in favor of #237, which updates the generation script to generate a CircleCI configuration that uses the current best practices of using the `.palantir/go-version` file to determine the current and previous Go version.

The advantage of the alternate approach is that it integrates nicely with the existing `manage-go-version-oss` check -- as can be seen by #238, with the updated configuration the standard check will properly update both the `go.mod` file and Go versions for all of the modules in the repository.

PR #252, opened 2022-06-01
break: refreshable/v2 with Generic type handling

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/252)


refreshable/refreshable.go
```diff
@@ -4,14 +4,91 @@
...
 package refreshable
 
-type Refreshable interface {
+import (
+	"context"
+)
+
+// A Refreshable is a generic container type for a volatile underlying value.
+// It supports atomic access and user-provided callback "subscriptions" on updates.
+type Refreshable[T any] interface {
```
ravi-braj, line 13: I would also be keen to see an implementation of this that allows clients to create refreshables that update periodically. Clients could provide a ticker duration & an update function.

[This is what I've been playing around with](https://github.palantir.build/deployability/cloud-provider-constructors/blob/d8a669e063b1b62aafa53ec83bc8d0ef3bd7352d/credential/periodic_refresher.go) on top of the non-generics implementation

bmoylan: @ravi-braj is this handled by `NewFromTickerFunc[T]` or are you looking for something else?

ravi-braj: The Ready type doesn't quite fit what I'm looking for as the values I have in mind will multiple times over their lifetime. The example I have is short lived credentials that need to be refreshed / re-read from disk every few minutes

bmoylan: NewFromTickerFunc will check for a new value every time the ticker fires. It is intentional to not return `Updatable` from the constructors that handle updates internally, since the caller should only be reading from the refreshable. The `ReadyC` channel allows you to block until a value is populated.

refreshable/async.go
```diff
@@ -0,0 +1,100 @@
...
+	return r.in.Subscribe(consumer)
+}
+
+func (r *ready[T]) ReadyC() <-chan struct{} {
+	return r.readyC
+}
+
+func (r *ready[T]) Update(val T) {
+	r.cancel()
+	r.in.Update(val)
```
bwaldrep, line 41: Shouldn't the order here be reversed?

bmoylan: yep, fixed, thanks!

refreshable/async.go
```diff
@@ -0,0 +1,100 @@
...
+			out.Update(value)
+		}
+	}()
+	return out
+}
+
+// NewFromTickerFunc returns a Ready Refreshable populated by the result of the provider called each interval.
+// If the providers bool return is false, the value is ignored.
+// The result's ReadyC channel is closed when a new value is populated.
+func NewFromTickerFunc[T any](interval time.Duration, provider func() (T, bool)) (Ready[T], UnsubscribeFunc) {
```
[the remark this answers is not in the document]
bmoylan: done!

k-simons:
👍

PR #268, opened 2022-09-20
Update nmiyake/pkg/{gofiles,dirs} to latest

tabboud:
Updating a few internal products that hit ambiguous imports due to the nmiyake/pkg tagging in v1.0.0

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/268)


tabboud:
closing in favor of removing the pkgs entirely

PR #318, opened 2023-08-15
Metrics concurrency bug

CoreyCook8:
There is a race condition under certain situations when using the FromContext function call and then updating metrics.

Each call to FromContext, when a tagsContainer is present, will return a childRegistry with the tags field set to the same location in memory.

In the case where these tag slices have a capacity > length and concurrent calls to append are made on the slice (Such as two processes calling Gauge at the same time), the tags slice will be populated with the first caller's new tags, and then overwritten by the second caller. If this happens before the tags are copied in registerMetric, the first caller will have the name of the first caller and the tags of the second caller.

In this test case, since we add tags to the context one by one for a total of 3. The tags slice will have a length of 3 and a capacity of 4. Because of that we see the behavior described above. 

Gauge is called with prefix1 and tag {"name": prefix1}. The append call does not create a new array since the memory is allocated to fit this 1 tag. Then, before the tags are copied, Gauge is called with prefix2 and tag {"name":prefix2}. The amend in this case overwrites the {"name":prefix1} since this is done on the same slice, and then by the time the prefix1 tags are being copied the contents contain {"name":prefix2} and not {"name":prefix1}.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/318)

### DIFF
--- a/metrics/context.go
+++ b/metrics/context.go
@@ -34,9 +34,14 @@ func FromContext(ctx context.Context) Registry {
 	if !ok {
 		return registry
 	}
+	var tags Tags
+	if len(tagsContainer.Tags) > 0 {
+		tags = make(Tags, len(tagsContainer.Tags))
+		copy(tags, tagsContainer.Tags)
+	}
 	return &childRegistry{
 		root: rootRegistry,
-		tags: tagsContainer.Tags,
+		tags: tags,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


metrics/context.go
```diff
@@ -52,8 +52,11 @@ func AddTags(ctx context.Context, tags ...Tag) context.Context {
 	if !ok || container == nil {
 		container = &tagsContainer{}
 	}
+	newTags := make(Tags, len(container.Tags), len(container.Tags)+len(tags))
```
bmoylan, line 55: Let's use copy for both input slices:
```go
newTags := make(Tags, len(container.Tags)+len(tags))
copy(newTags, container.Tags)
copy(newTags[len(container.Tags):], tags)
```

bmoylan (approved):
Thanks for the root-causing and contribution!

PR #349, opened 2024-08-27
Add ability to patch yaml with comments

Arthur-Befumo:
This PR updates the `Patch` struct to include a `Comment` field, which is piped to the `HeadComment` field in the generated `yaml.Node` for `Add` and `Replace` operations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/349)

### DIFF
--- a/yamlpatch/patch.go
+++ b/yamlpatch/patch.go
@@ -34,17 +34,17 @@ func Apply(originalBytes []byte, patch Patch) ([]byte, error) {
 		var err error
 		switch op.Type {
 		case OperationAdd:
-			err = patchAdd(node, op.Path, op.Value)
+			err = patchAdd(node, op.Path, op.Value, op.Comment)
 		case OperationRemove:
 			err = patchRemove(node, op.Path)
 		case OperationReplace:
-			err = patchReplace(node, op.Path, op.Value)
+			err = patchReplace(node, op.Path, op.Value, op.Comment)
 		case OperationMove:
 			err = patchMove(node, op.Path, op.From)
 		case OperationCopy:
 			err = patchCopy(node, op.Path, op.From)
 		case OperationTest:
-			err = patchTest(node, op.Path, op.Value)
+			err = patchTest(node, op.Path, op.Value, op.Comment)
 		default:
 			err = errors.Errorf("unexpected op")
 		}
@@ -64,8 +64,8 @@ func Apply(originalBytes []byte, patch Patch) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-func patchAdd(node *yaml.Node, path Path, value interface{}) error {
-	valueNode, err := valueToYAMLNode(value)
+func patchAdd(node *yaml.Node, path Path, value interface{}, comment string) error {
+	valueNode, err := valueToYAMLNode(value, comment)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


yamlpatch/types.go
```diff
@@ -28,6 +28,9 @@ type Operation struct {
 	Path  Path        `json:"path" yaml:"path"`
 	From  Path        `json:"from,omitempty" yaml:"from,omitempty"`
 	Value interface{} `json:"value,omitempty" yaml:"value,omitempty"`
+
+	// If not empty, will be inserted as a head comment above this node
```
bmoylan, line 32: Can you include which operations support adding the comment? Looks like add/replace/test. Should we error for the others if the incorrectly specify a comment?

Maybe also worth mentioning that this is not part of RFC6902

yamlpatch/patch_test.go
```diff
@@ -82,6 +82,26 @@ foo:
 foo:
   arr: [1, 2, 3, 4]
   bar: 1 # my bar`,
+		},
+		{
+			Name:  "add to array with comment",
```
bmoylan, line 87: can we also add a test for an object/map and the test operation?

yamlpatch/patch.go
```diff
@@ -232,6 +232,7 @@ func valueToYAMLNode(value interface{}) (*yaml.Node, error) {
 		return nil, err
 	}
 	clearYAMLStyle(node)
+	node.HeadComment = comment
```
bmoylan, line 235: Will this overwrite an existing comment every time? Should this be conditioned on whether the `comment == ""`? If so, do we also need a way to _remove_ an existing comment?

bmoylan (changes requested):

yamlpatch/patch.go
Arthur-Befumo: This node is constructed by marshaling the value (which is `interface{}`) into yaml, and then unmarshalling that into a `yaml.Node` ... so prior to this PR, the `node.HeadComment` field would always be empty (no change in behavior for operations that do not define a comment). 

When a comment is included it gets inserted above the value and won't impact other comments in the original document that are not in the scope of that node.

Example of behavior doing a `Replace` on `foo/bar` path with a comment in the operation:
```
foo:
  # some other comment
  bar: # inline comment
  # original comment
  - 10
  - 20
```

```
foo:
  # some other comment
  bar: # inline comment
  # patched comment
  - 10
  - 20
```
vs prior to this PR or without a comment (equivalent):
```
foo:
  # some other comment
  bar: # inline comment
  - 10
  - 20
```

yamlpatch/patch_test.go
Arthur-Befumo: Added a test for `replace` with an object

I looked more at the test operation and realized that the comment isn't relevant / never gets consumed there, so I've stopped passing the comment to that operation entirely.

yamlpatch/patch_test.go
```diff
@@ -94,6 +94,50 @@ foo:
...
+foo:
+  bar: hello-world # previous comment`,
+			Expected: `# my foo
+foo:
+  bar:
+    # key value pair
+    key: value`,
+		},
+		{
+			Name:  "test object with comment",
```
bmoylan, line 131: looks like this is the same test as above?

Arthur-Befumo: Whoops, fixed

bmoylan (approved):

PR #359, opened 2025-01-17
refreshables: New constructors for more flexibility in how derivative refreshables are created

bmoylan:
- `Cached()` subscribes to an upstream refreshable and stores the result in one it creates. 
- `View()` does not create a subscription nor a cached value of the resulting type. Instead, the implementations of `Current()` and `Subscribe()` convert the upstream Refreshable's return values on every invocation. This reduces memory pressure for cheap or infrequent conversions like struct field accessors.
- `Merge()` combines two refreshables into one using a `reduceFn`. An update to either will refresh the reduced value.
- `Collect()` combines a list of refreshables of the same type into a refreshable slice type. 

This PR also adds some `Example*` tests to improve the package's godoc.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/359)

### DIFF
--- a/refreshable/refreshable.go
+++ b/refreshable/refreshable.go
@@ -112,38 +112,38 @@ func Validate[T any](original Refreshable[T], validatingFn func(T) error) (Valid
 	return MapWithError(original, identity(validatingFn))
 }
 
-// Reduce returns a new Refreshable that combines the latest values of two Refreshables using the reduceFn.
+// Merge returns a new Refreshable that combines the latest values of two Refreshables of different types using the reduceFn.
 // The returned Refreshable is updated whenever either of the original Refreshables updates.
-// The unsubscribe function removes subscriptions from both initial Refreshables.
-func Reduce[T1 any, T2 any, R any](reduceFn func(T1, T2) R, initial1 Refreshable[T1], initial2 Refreshable[T2]) (Refreshable[R], UnsubscribeFunc) {
+// The unsubscribe function removes subscriptions from both original Refreshables.
+func Merge[T1 any, T2 any, R any](original1 Refreshable[T1], original2 Refreshable[T2], reduceFn func(T1, T2) R) (Refreshable[R], UnsubscribeFunc) {
 	out := newZero[R]()
-	stop1 := initial1.Subscribe(func(v T1) {
-		out.Update(reduceFn(v, initial2.Current()))
-	})
-	stop2 := initial2.Subscribe(func(v T2) {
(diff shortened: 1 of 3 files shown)


refreshable/refreshable.go
```diff
@@ -93,3 +111,43 @@ func MapWithError[T any, M any](original Refreshable[T], mapFn func(T) (M, error
...
+		stop1()
+		stop2()
+	}
+}
+
+// ReduceN returns a new Refreshable that combines the latest values of multiple Refreshables using the reduceFn.
+// The reduceFn is called with a slice of the current values of the original Refreshables.
+// The returned Refreshable is updated whenever any of the original Refreshables updates.
+// The unsubscribe function removes subscriptions from all initial Refreshables.
+func ReduceN[T any, R any](reduceFn func([]T) R, initial ...Refreshable[T]) (Refreshable[R], UnsubscribeFunc) {
```
glynternet, line 136: Shall we rename to `Reduce` and remove the original `Reduce`?
`ReduceN` can be used for 2 elements already.

I think most collection Reduce functions have a signature that takes an accumulator and an initial value, so with T and R here it would be like:
```go
func Reduce(accumulator func(element T, accumulation R), initial R, refreshables ...Refreshable[T]) (Refreshable[R], UnsubscribeFunc)
```

I'm not against us going for the format here, tbh, but I do think we should at least consider that we are diverging from the norm.

Are we also at the point where we want to accept an `iter.Seq[Refreshable[T]]` instead of `...Refreshable[T]`?
I haven't come across any uses of `iter.Seq` yet but maybe it would be bad to get trapped with `...Refreshable[T]` becomes the preferred method of supplying ranges/collections in the future 🤷

bmoylan: Ok I pushed a couple renames that might be better...
* `Reduce` > `Merge`: The important thing about that one is the input refreshables are two different types, unlike the list/slice/variadic one.
* `ReduceN` > `Collect`, and no longer take a function; just return a `Refreshable[[]T]` and if the caller wants to map that to something else, they can.

bmoylan: On the iter.Seq thing, I haven't seen a lot of APIs support it either, but in this case they could do `refreshables.Collect(slices.Collect(seq)...)` so I'm not _that_ worried while we see how the best practices evolve

glynternet: Nice, appreciated!

nmiyake (approved):

nmiyake (approved):

PR #373, opened 2025-03-27
Add metrics.CaptureRuntimeMemStatsFunc()

bmoylan:
Allows for calling CaptureRuntimeMemStats explicitly rather than wrapping it in its own goroutine/ticker.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/373)

### DIFF
--- a/metrics/registry.go
+++ b/metrics/registry.go
@@ -147,22 +147,24 @@ func CaptureRuntimeMemStatsWithContext(ctx context.Context, registry RootRegistr
 	return true
 }
 
-// CaptureRuntimeMemStatsOnce captures the runtime memory statistics once and registers them in the provided registry.
+// CaptureRuntimeMemStatsFunc initializes the runtime metric gauges and returns a function that captures the runtime
+// memory statistics once and registers them in the provided registry.
 // This function only supports RootRegistry implementations that implement the metricsRegistryProvider interface --
 // if the provided RootRegistry does not satisfy this interface, this function is a no-op. This function returns true
 // if it starts the runtime metric collection goroutine, false otherwise.
 //
 // The gauges/metrics etc. used to track runtime statistics are shared globally and the values are reset every time this
 // function is called (if it is not a no-op).
-func CaptureRuntimeMemStatsOnce(registry RootRegistry) bool {
+func CaptureRuntimeMemStatsFunc(registry RootRegistry) (func(), bool) {
 	mRegProvider, ok := registry.(metricsRegistryProvider)
 	if !ok {
-		return false
+		return nil, false
 	}
 	goRegistry := metrics.NewPrefixedChildRegistry(mRegProvider.Registry(), "go.")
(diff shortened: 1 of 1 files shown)


metrics/registry.go
```diff
@@ -147,6 +147,24 @@ func CaptureRuntimeMemStatsWithContext(ctx context.Context, registry RootRegistr
...
+func CaptureRuntimeMemStatsOnce(registry RootRegistry) bool {
+	mRegProvider, ok := registry.(metricsRegistryProvider)
+	if !ok {
+		return false
+	}
+	goRegistry := metrics.NewPrefixedChildRegistry(mRegProvider.Registry(), "go.")
+	metrics.RegisterRuntimeMemStats(goRegistry)
+	metrics.CaptureRuntimeMemStatsOnce(goRegistry)
+	return true
+}
```
nmiyake, line 166: This function feels a bit awkward -- presumably it exists so that it can be called repeatedly/multiple times, but the entire highlighted portion here is only really meant to be run once. Also, in the PR where this is proposed to be used, the return value doesn't seem to be checked at all.

It seems like the intention here is that you want to be able to capture the runtime stats together with other stats in a routine you control. In that case, I would recommend having this function do the setup and then return a function that can be invoked to capture the stats on a one-off basis, and then call that function in your routine:

```suggestion
func CaptureRuntimeMemStatsOnce(registry RootRegistry) (fun(), bool) {
	mRegProvider, ok := registry.(metricsRegistryProvider)
	if !ok {
		return nil, false
	}
	goRegistry := metrics.NewPrefixedChildRegistry(mRegProvider.Registry(), "go.")
	metrics.RegisterRuntimeMemStats(goRegistry)
	return func() {
	      metrics.CaptureRuntimeMemStatsOnce(goRegistry)
	}, return true
}
```

This way, the initialization still only happens once, but the caller has a way to make the one-off "capture" calls as they see fit.

nmiyake (changes requested):

metrics/registry.go
bmoylan: Great suggestion! I've updated this function and renamed it `CaptureRuntimeMemStatsFunc` now that it returns a function

metrics/registry.go
```diff
@@ -147,6 +147,26 @@ func CaptureRuntimeMemStatsWithContext(ctx context.Context, registry RootRegistr
 	return true
 }
 
+// CaptureRuntimeMemStatsFunc initializes the runtime metric gauges and returns a function that captures the runtime
+// memory statistics once and registers them in the provided registry.
```
nmiyake, line 151:
```suggestion
// CaptureRuntimeMemStatsFunc initializes the runtime metric gauges and registers them in the provided registry and returns a function that captures the runtime statistics when called.
```

nmiyake (approved):
Minor feedback on comment/documentation for function, but non-blocking so providing approval.

nmiyake (approved):

PR #374, opened 2025-03-27
metrics.MetricVal: Add Keys() and Value() to reduce allocations when counting cardinality and emitting metric logs

bmoylan:
witchcraft-go-server is seeing a lot of allocation activity calculating metric values that are blacklisted/skipped. By deferring their computation per-key, we can avoid that wasted work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/374)

### DIFF
--- a/metrics/value.go
+++ b/metrics/value.go
@@ -10,6 +10,21 @@ import (
 
 type MetricVal interface {
 	Type() string
+
+	// Keys implements iter.Seq[string] by returning the keys that can be used to retrieve values from the metric.
+	// Example:
+	//
+	//	for key := range val.Keys {
+	//		if !skip(key) {
+	//			fmt.Println(key, mv.Value(key))
+	// 		}
+	//	}
+	Keys(yield func(string) bool)
+
+	// Value returns the computed value for the given key. If the key is not recognized, returns nil.
+	Value(key string) interface{}
+
+	// Deprecated: use Keys and Value to iterate through values and avoid eager computation of skipped keys.
 	Values() map[string]interface{}
 }
 
@@ -39,6 +54,22 @@ func (v *counterVal) Type() string {
 	return "counter"
 }
 
+func (v *counterVal) Keys(yield func(string) bool) {
+	for _, key := range []string{"count"} {
+		if !yield(key) {
+			return
+		}
+	}
+}
+
+func (v *counterVal) Value(key string) interface{} {
+	switch key {
+	case "count":
+		return v.Counter.Count()
+	}
+	return nil
+}
+
 func (v *counterVal) Values() map[string]interface{} {
 	return map[string]interface{}{
 		"count": v.Count(),
@@ -53,9 +84,26 @@ func (v *gaugeVal) Type() string {
 	return "gauge"
 }
 
+func (v *gaugeVal) Keys(yield func(string) bool) {
+	for _, key := range []string{"value"} {
+		if !yield(key) {
+			return
+		}
+	}
+}
+
+func (v *gaugeVal) Value(key string) interface{} {
+	switch key {
+	case "value":
+		return v.Gauge.Value()
+	default:
+		return nil
+	}
+}
+
 func (v *gaugeVal) Values() map[string]interface{} {
 	return map[string]interface{}{
-		"value": v.Value(),
+		"value": v.Gauge.Value(),
 	}
 }
 
@@ -67,9 +115,26 @@ func (v *gaugeFloat64Val) Type() string {
 	return "gauge"
 }
 
+func (v *gaugeFloat64Val) Keys(yield func(string) bool) {
+	for _, key := range []string{"value"} {
+		if !yield(key) {
+			return
+		}
+	}
+}
+
+func (v *gaugeFloat64Val) Value(key string) interface{} {
+	switch key {
+	case "value":
+		return v.GaugeFloat64.Value()
+	default:
+		return nil
+	}
+}
+
 func (v *gaugeFloat64Val) Values() map[string]interface{} {
 	return map[string]interface{}{
-		"value": v.Value(),
+		"value": v.GaugeFloat64.Value(),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


metrics/value.go
```diff
@@ -10,6 +10,21 @@ import (
...
+
+	// Keys implements iter.Seq[string] by returning the keys that can be used to retrieve values from the metric.
+	// Example:
+	//
+	//	for key := range val.Keys {
+	//		if !skip(key) {
+	//			fmt.Println(key, mv.Value(key))
+	// 		}
+	//	}
+	Keys(yield func(string) bool)
```
nmiyake, line 22: Zooming out: any reason not to just have `Keys() []string` here rather than implementing/returning an iterator? The implementation of the iterator basically allocates a static string slice and then iterates through that and the only proposed usage I see iterates over all elements of the iterator using a `range` statement, so not sure that using an iterator adds much in this case.

*If* we decide that we want to use an iterator here, for conceptual clarity and to match Go best practices, let's have this function return an iterator rather than be an iterator itself:

```suggestion
     Keys() iter.Seq[string]
```

Per the Go blog entry (https://go.dev/blog/range-functions):

> In this kind of code s.All is a method that returns a function. We are calling s.All, and then using for/range to range over the function that it returns. In this case we could have made Set.All be an iterator function itself, rather than having it return an iterator function. However, in some cases that won’t work, such as if the function that returns the iterator needs to take an argument, or needs to do some set up work. As a matter of convention, we encourage all container types to provide an All method that returns an iterator, so that programmers don’t have to remember whether to range over All directly or whether to call All to get a value they can range over. They can always do the latter.

I also think the "this function returns an iterator" signature is easier to reason about compared to "this function itself is the iterator" (case in point, I had to look up/do background reading on this to reacquaint myself on this :)

metrics/value.go
```diff
@@ -39,6 +54,22 @@ func (v *counterVal) Type() string {
 	return "counter"
 }
 
+func (v *counterVal) Keys(yield func(string) bool) {
+	for _, key := range []string{"count"} {
+		if !yield(key) {
+			return
+		}
+	}
+}
```
nmiyake, line 63: The various `keys` implementation repeats a lot of boilerplate when the only difference is the input keys.

If we go with returning string slices for keys, let's just return slices directly.

If we go with returning iterators, would recommend having a helper function and then implementing various `Keys()` implementations as follows:

```suggestion
func stringSeq(strings ...string) iter.Seq[string] {
	return func(yield func(string) bool) {
		for _, key := range strings {
			if !yield(key) {
				return
			}
		}
	}
}

func (v *counterVal) Keys() iter.Seq[string] {
    return stringSeq("count")
}
```

metrics/value.go
```diff
@@ -53,9 +84,26 @@ func (v *gaugeVal) Type() string {
...
+
+func (v *gaugeVal) Value(key string) interface{} {
+	switch key {
+	case "value":
+		return v.Gauge.Value()
+	default:
+		return nil
+	}
```
nmiyake, line 101: Minor, but style/structure differs from `*counterVal` implementation -- let's choose either "return nil outside of loop" or "return nil in default case" and use same approach for all implementations

metrics/value.go
```diff
@@ -53,9 +84,26 @@ func (v *gaugeVal) Type() string {
...
+}
+
 func (v *gaugeVal) Values() map[string]interface{} {
 	return map[string]interface{}{
-		"value": v.Value(),
+		"value": v.Gauge.Value(),
```
nmiyake, line 106: This is a functional change -- is it intentional? Seems correct, but wanted to verify.

nmiyake: Related: can we update all the existing `Values` implementations to use the new functions instead? Basically, allocate a map, iterate over all values returned by `keys()` and then assign the value returned by `Value()` (can have a helper function that does this and just call it from all implementations)?

Although I know that `Values()` is now deprecated and unlikely to be used in the future, would like it to share the same underlying source of truth as `Keys()/Value()` combination to avoid any bugs/issues where we update a key or value implementation in one place but not the other.

nmiyake (changes requested):

metrics/value.go
bmoylan: Yeah I was concerned about allocating the anonymous function on each call, but I've run the benchmarks and it looks like it gets optimized away. Updated to `Keys() iter.Seq[string]`.

metrics/value.go
bmoylan: :+1: thanks, fixed

metrics/value.go
bmoylan: I attempted the `stringSeq` helper, but it allocates the slice of arguments on each call so I'm not sure it's worth it.

metrics/value.go
bmoylan: This is equivalent because previously `v.Value()` was calling the embedded metric type's `Value()` method, but since `*gaugeVal` now has a `Value(string)` that no longer works.

I've implemented your suggestion to implement `Values()` based on `Keys()/Value()` in a helper `collectValuesByKey` 👍

metrics/value.go
nmiyake: OK, thanks for this context.

Question, though -- doesn't the current implementation allocate a new slice on each call as well? For code like:

```
func (v *histogramVal) Keys() iter.Seq[string] {
	return func(yield func(string) bool) {
		for _, key := range []string{"count", "min", "max", "mean", "stddev", "p50", "p95", "p99"} {
```

Unless the compiler is doing optimization under the hood (which it might be), it seems like each call to `Keys()` would create a new anonymous function that allocates a new static slice to range over, which wouldn't seem to be much different than returning the slice directly. If the allocation is something we need to be worried about, shouldn't this code define a private static slice for each of these key sets and range over that? This isn't a major issue either way, but just wanted to be sure I understood the concern and verify that current code addresses it (I'm also fine with approving current approach -- usability between iterator and returning keys directly is really not any different)

metrics/value.go
```diff
@@ -39,12 +56,29 @@ func (v *counterVal) Type() string {
...
-	return map[string]interface{}{
-		"count": v.Count(),
+func (v *counterVal) Keys() iter.Seq[string] {
+	return func(yield func(string) bool) {
+		for _, key := range []string{"count"} {
+			if !yield(key) {
+				return
+			}
+		}
+	}
```
nmiyake, line 66: Seems like we could decomp out a function that takes the slice of keys and then returns the iterator function.

If each argument slice is a local variable (`counterValKeys = []string{"count"}`, etc.), then calling the helper function should just be a reference to the slice (so no extra allocation/copying), and should also avoid allocation on each iterator call, right? If so, seems like this would be strictly better both from performance and reducing duplicated code.

nmiyake:
Minor feedback, but non-blocking/optional so providing approval.

nmiyake (approved):

metrics/value.go
bmoylan: Yeah, when I build with the memory debug flags, I see the compiler deciding `value.go:174:31: []string{...} does not escape` and it doesn't actually allocate the slice. It does have to allocate the individual string values though.

metrics/value.go
bmoylan: I think you're basically describing [`slices.Values`](https://pkg.go.dev/slices#Values). I tested this and it results in an extra allocation (I think because the slice has to go on the heap to be passed into that function)

```

```

nmiyake:
Thanks for all the investigating! All my concerns are addressed so fully lgtm

PR #410, opened 2026-01-07
✨ Add PrintInfoLevelErrorAndParamsWithDebugTransformer error handler

glynternet:
Add a cobra error handler that prints error messages along with werror params (both safe and unsafe). When debug mode is enabled, delegates to a custom transformer function instead for detailed error output.

## Changes Made
- **New function `PrintInfoLevelErrorAndParamsWithDebugTransformer`**: Cobra error handler that extracts and prints werror safe/unsafe params alongside error messages
- **New helper `formattedParamLine`**: Formats individual param key-value pairs with JSON marshalling
- **New dependency**: Added `github.com/palantir/witchcraft-go-error` for werror param extraction

## Risk Assessment

**Risk Level**: Low ⚡
- Adds new functionality without modifying existing behavior
- All existing tests continue to pass
- New function has comprehensive test coverage

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/410)

### DIFF
--- a/cobracli/execute.go
+++ b/cobracli/execute.go
@@ -5,10 +5,15 @@
 package cobracli
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"sort"
 	"strings"
 
+	werror "github.com/palantir/witchcraft-go-error"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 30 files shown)


cobracli/execute.go
```diff
@@ -205,3 +210,64 @@ func FlagErrorsUsageErrorConfigurer(command *cobra.Command) {
 		return fmt.Errorf("%s\n%s", err.Error(), strings.TrimSuffix(c.UsageString(), "\n"))
 	})
 }
+
+// PrintInfoLevelErrorAndParamsWithDebugTransformer provides a cobra error handler that will print the error message and params as standard.
```
nmiyake: > as standard.

A bit confused by this wording -- does it mean as a standard part of the operation? Or to standard out?

cobracli/execute.go
```diff
@@ -205,3 +210,64 @@ func FlagErrorsUsageErrorConfigurer(command *cobra.Command) {
...
+		if err == nil || err.Error() == "" {
+			return
+		}
+
+		var buf bytes.Buffer
+		// always print contents of buffer no matter what return path from function
+		defer func() {
+			if _, err := buf.WriteTo(command.ErrOrStderr()); err != nil {
+				_, _ = fmt.Fprintln(os.Stdout, "Error copying error buffer to file:", err.Error())
+				_, _ = fmt.Fprintln(os.Stdout, "Buffer:", buf.String())
```
nmiyake: Wondering if it make more sense to have the function take an optional `errWriter` instead (and maybe default to `os.Stderr` if that is `nil`)?

It might be a bit heavy-handed to dictate that error writing to the output stream for the command will force error output to `os.Stdout`

cobracli/execute.go
```diff
@@ -205,3 +210,64 @@ func FlagErrorsUsageErrorConfigurer(command *cobra.Command) {
...
+		if len(safe) == 0 && len(unsafe) == 0 {
+			return
+		}
+		var keys []string
+		for key := range safe {
+			keys = append(keys, key)
+		}
+		for key := range unsafe {
+			keys = append(keys, key)
+		}
```
nmiyake: nit, but I think this is a bit more readable:

```suggestion
		keys = slices.AppendSeq(keys, maps.Keys(safe))
		keys = slices.AppendSeq(keys, maps.Keys(unsafe))
```

cobracli/execute.go
```diff
@@ -205,3 +210,64 @@ func FlagErrorsUsageErrorConfigurer(command *cobra.Command) {
...
+			}
+			if val, ok := unsafe[key]; ok {
+				// writes to bytes.Buffer always return nil error
+				_, _ = fmt.Fprintln(&buf, formattedParamLine(key, val))
+			}
+		}
+	}
+}
+
+func formattedParamLine(key string, val interface{}) string {
```
nmiyake:
```suggestion
func formattedParamLine(key string, val any) string {
```

cobracli/execute.go
nmiyake: Also, I think it's worth mentioning in the documentation that the default implementation will print the JSON-marshaled representation of values, or the error encountered while trying to marshal the value as JSON

nmiyake (changes requested):
Thanks! Some feedback.

cobracli/execute.go
glynternet: Oh wow, yeah, this code has followed me around for years, I guess I never got around to updating to modern go std pkg APIs. Thx.

nmiyake (approved):

PR #437, opened 2026-03-26
refreshable: replace interface{} -> any

nmiyake:
Runs "go fix" on the refreshable package. The main change is that the references to "interface{}" are replaced with "any".

Even though this does not make a functional difference, it is helpful for repositories that generate code based on the signatures in this package, as it helps the output to remain compatible with "go fix".

## After this PR

==COMMIT_MSG==
Replaces all instances of `interface{}` with `any`.
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/pkg/437)

### DIFF
--- a/refreshable/main.go
+++ b/refreshable/main.go
@@ -3,7 +3,6 @@
 // license that can be found in the LICENSE file.
 
 //go:build module
-// +build module
 
 // This file exists only to smooth the transition for modules. Having this file makes it such that other modules that
 // consume this module will not have import path conflicts caused by github.com/palantir/pkg.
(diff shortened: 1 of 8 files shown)


nmiyake:
`refreshable/v1` is the long-running feature branch for the `v1` version of the module.

This will be released as `refreshable/v1.8.0`. This will also help to ensure that the latest release on the v1 branch is correct, as [v1.7.0 was incorrectly tagged on the v2 module](https://github.com/palantir/pkg/releases/tag/refreshable%2Fv1.7.0).

tabboud (approved):

